PR #123, opened 2023-11-20
feature: create seamless tileable images in x or y dimension

brycedrennan:
Adds a seamless tile mode but does so in a way you guys probably won't like :-)  Opening the pull request to have the discussion.

Tile mode can be set to tile in both directions or only "x" or "y".  Set an image as desktop background in tile mode to really get the full effect.

I suspect this will work with SDXL as well but haven't tried it yet.

Examples:
022785_781162360_kdpmpp2m15_PS7 5_white_marble_quartz_with_gold_veins_ generated
284143311-fefd356d-3050-4dfc-89a6-3f30ae8e91ab_reduced_perfect_loop

tiled_image_1_2x3

022755_297564700_kdpmpp2m35_PS7 5_blueberries_ generated
284144042-06c64067-aa5d-4f9f-8b50-d691200c4f75_reduced_perfect_loop
tiled_image_2_2x3

022559_809130920_kdpmpp2m35_PS7 5_an_oil_drop_ generated
022559_809130920_kdpmpp2m35_PS7 5_an_oil_drop_ generated _reduced_perfect_loop
tiled_image_2x3

### DIFF
--- a/src/refiners/fluxion/layers/chain.py
+++ b/src/refiners/fluxion/layers/chain.py
@@ -257,6 +257,7 @@ class Chain(ContextModule):
         try:
             return layer(*args)
         except Exception as e:
+            raise e
             exc_type, _, exc_traceback = sys.exc_info()
             assert exc_type
             tb_list = traceback.extract_tb(tb=exc_traceback)
--- a/src/refiners/foundationals/latent_diffusion/seamless_tiles.py
+++ b/src/refiners/foundationals/latent_diffusion/seamless_tiles.py
@@ -1,0 +1,88 @@
+"""Code for enabling single direction circular padding on Conv2d layers."""
+
+from typing import Literal
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+
+import refiners.fluxion.layers as fl
+from refiners.fluxion.adapters import Adapter
+from refiners.foundationals.latent_diffusion.stable_diffusion_1.unet import SD1UNet
+
+TileModeType = Literal["", "x", "y", "xy"]
+PaddingModeType = Literal["circular", "constant"]
+
+
+class SeamlessConv2dWrapper(fl.Module):
+    def __init__(self, target: fl.Conv2d, tile_adapter: "TilingAdapter") -> None:
+        super().__init__()
+        self.target = target
+        self.tile_adapter = tile_adapter
+
+    def _conv_forward(self, input: torch.Tensor, weight: torch.Tensor, bias: torch.Tensor) -> torch.Tensor:
+        tile_mode = self.tile_adapter.tile_mode
+        padding_mode_x = "circular" if "x" in tile_mode else "constant"
+        padding_mode_y = "circular" if "y" in tile_mode else "constant"
+
+        if padding_mode_y == padding_mode_x:
+            original_padding_mode = self.target.padding_mode
+            try:
+                self.target.padding_mode = padding_mode_x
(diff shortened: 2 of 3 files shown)


brycedrennan:
Also FYI diffusers open feature request for the same feature: https://github.com/huggingface/diffusers/issues/556

limiteinductive:
It's a great idea! We generally don't directly modify the existing implementation of UNet but instead work with the [Adapter pattern ](https://github.com/finegrain-ai/refiners/blob/main/src/refiners/fluxion/adapters/adapter.py#L10)([e.g.](https://github.com/finegrain-ai/refiners/blob/main/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/t2i_adapter.py#L8)

-> You first write a TilingConv2dAdapter that targets a fl.Conv2d that wraps the original conv and modifies it.
-> You then write a TilingAdapter that targets UNets and inject TilingConv2dAdapter into all their conv.

brycedrennan:
Great thank you- I'll give it a go.

brycedrennan:
Was revisiting this. Should `TilingConv2dAdapter` inherit from `Adapter`?  I ask because [this assert](https://github.com/finegrain-ai/refiners/blob/e7892254eb85b06212515b5482a4fa3f699650f8/src/refiners/fluxion/adapters/adapter.py#L24)  seems to imply that an Adapter cannot wrap `fl.Conv2d` since it's not a chain

catwell:
@brycedrennan Adapters can wrap `Conv2d`, they cannot be a Conv2d (they must subclass Chain). `self` is the adapter here, the conv would be the adaptee (the generic `T`).

deltheil:
> Adapters can wrap `Conv2d`

@brycedrennan FYI, you have such an example part of the UNet: https://github.com/finegrain-ai/refiners/blob/e7892254eb85b06212515b5482a4fa3f699650f8/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/unet.py#L263-L270

brycedrennan:
Ah - i see thank you

brycedrennan:
Well I'm pretty sure this is wrong in multiple ways, but it does work.  I wasn't quite sure what the "right" way was to wrap the conv2d.

Additionally, it runs roughly .6 it/s slower than the method in the first commit that does not use adapters

src/refiners/fluxion/layers/chain.py
```diff
@@ -257,6 +257,7 @@ def _call_layer(self, layer: Module, name: str, /, *args: Any) -> Any:
         try:
             return layer(*args)
         except Exception as e:
+            raise e
```
brycedrennan, line 260: whoops. remove.

deltheil:
> Well I'm pretty sure this is wrong in multiple ways, but it does work.

Hey @brycedrennan, do you consider it "ready for review" or not?

brycedrennan:
Sorry to have let this sit so long.  The proposed changes work but are slower than the implementation that directly patches conv2d.  I'm not sure why that is but doesn't feel right to propose merging as is. 

Since the feature is already working in imaginairy there isn't much urgency on my part.  I'd be okay with this pull request being closed or alternatively if you have feedback on it I can continue iterating on it.

limiteinductive:
Maybe what could be done is just having an Adapter on UNet that would patch all Conv2d as a whole instead of wrapping each of them; that would work, too.

brycedrennan:
okay I'll try that thanks

PR #130, opened 2023-12-01
feature: add sliced-attention for memory efficiency

brycedrennan:
Setting slice-size to 4096 allowed me to produce HD images on M1 32gb and 7000x5000 on Nvidia 4090

I saw no visual difference in images generated.

4k image produced:
`imagine "abstract oil painting. moder art" -w 3840 -h 2160 --steps 50 --sampler ddim`
023112_461835499_ddim50_PS7 5_abstract_oil_painting _moder_art_ generated

### DIFF
--- a/src/refiners/fluxion/layers/attentions.py
+++ b/src/refiners/fluxion/layers/attentions.py
@@ -1,4 +1,5 @@
 import math
+
 import torch
 from jaxtyping import Float
 from torch.nn.functional import scaled_dot_product_attention as _scaled_dot_product_attention  # type: ignore
@@ -37,16 +38,52 @@ def sparse_dot_product_attention_non_optimized(
 
 
 class ScaledDotProductAttention(Module):
-    def __init__(self, num_heads: int = 1, is_causal: bool | None = None, is_optimized: bool = True) -> None:
+    def __init__(
+        self,
+        num_heads: int = 1,
+        is_causal: bool | None = None,
+        is_optimized: bool = True,
+        slice_size: int | None = None,
+    ) -> None:
         super().__init__()
         self.num_heads = num_heads
         self.is_causal = is_causal
         self.is_optimized = is_optimized
+        self.slice_size = slice_size
         self.dot_product = (
             scaled_dot_product_attention if self.is_optimized else sparse_dot_product_attention_non_optimized
         )
 
     def forward(
+        self,
+        query: Float[Tensor, "batch num_queries embedding_dim"],
+        key: Float[Tensor, "batch num_keys embedding_dim"],
+        value: Float[Tensor, "batch num_values embedding_dim"],
+        is_causal: bool | None = None,
+    ) -> Float[Tensor, "batch num_queries dim"]:
+        if self.slice_size is None:
+            return self._process_attention(query, key, value, is_causal)
+
+        return self._sliced_attention(query, key, value, is_causal=is_causal, slice_size=self.slice_size)
+
+    def _sliced_attention(
+        self,
+        query: Float[Tensor, "batch num_queries embedding_dim"],
+        key: Float[Tensor, "batch num_keys embedding_dim"],
+        value: Float[Tensor, "batch num_values embedding_dim"],
+        slice_size: int,
+        is_causal: bool | None = None,
+    ) -> Float[Tensor, "batch num_queries dim"]:
+        _, num_queries, _ = query.shape
+        output = torch.zeros_like(query)
+        for start_idx in range(0, num_queries, slice_size):
+            end_idx = min(start_idx + slice_size, num_queries)
+            output[:, start_idx:end_idx, :] = self._process_attention(
+                query[:, start_idx:end_idx, :], key, value, is_causal
+            )
+        return output
+
+    def _process_attention(
         self,
         query: Float[Tensor, "batch num_queries embedding_dim"],
         key: Float[Tensor, "batch num_keys embedding_dim"],


src/refiners/fluxion/layers/attentions.py
```diff
@@ -37,11 +39,24 @@ def sparse_dot_product_attention_non_optimized(
 
 
 class ScaledDotProductAttention(Module):
-    def __init__(self, num_heads: int = 1, is_causal: bool | None = None, is_optimized: bool = True) -> None:
+    def __init__(
+        self,
+        num_heads: int = 1,
+        is_causal: bool | None = None,
+        is_optimized: bool = True,
+        slice_size: Optional[int] = 4096,
```
deltheil, line 47: Since Refiners targets Python >=3.10, this should be `slice_size: int | None`.

deltheil, line 47: To keep current behavior unchanged, I would be inclined to use `None` as default:

```python
slice_size: int | None = None,
```

That being said: since `ScaledDotProductAttention` is deeply nested, how do you currently proceed if you want to change this advanced setting? (I mean there is no such option exposed at a high-level, like UNet CTOR)

Is it like you dynamically change it before hand? e.g. via `unet.layers(ScaledDotProductAttention)` and attribute update

deltheil (commented):
Thanks!

src/refiners/fluxion/layers/attentions.py
limiteinductive: It depends a lot if we consider sliced attention a general tool or specific to SD. I think it can be helpful in other contexts as well. Adding a CTOR kwarg for it wouldn't bother me (and a setter too).

src/refiners/fluxion/layers/attentions.py
```diff
@@ -37,11 +39,24 @@ def sparse_dot_product_attention_non_optimized(
...
         super().__init__()
         self.num_heads = num_heads
         self.is_causal = is_causal
         self.is_optimized = is_optimized
+        # 4096 max needed for SD 1.5 512x512
+        # 9216 max needed for SD 1.5 768x768
+        # 16384 max needed for SD 1.5 1024x1024
+        # 32400 max needed for SD 1.5 1920x1080 (HD)
+        # 129600 max needed for SD 1.5 3840x2160 (4k)
+        # 234375 max needed for SD 1.5 5000x3000
```
limiteinductive, line 58: `ScaledDotProductAttention` shouldn't "know" that SD exists, maybe you can document this in the `SD1UNet` docstring

brycedrennan: True, and I wasn't sure where this should be documented, I would have added some relevant SDXL values as well if I had that implemented. I just used it to inform the default I picked.  Also feels weird for the unet to have this comment.  I'd be okay just deleting it.  Perhaps it fits best in the `_sliced_attention` function.

src/refiners/fluxion/layers/attentions.py
brycedrennan: Also my not rigorous tests (I eyeballed two images of same seed) showed zero pixel differences in the output.

brycedrennan: In general I like to have defaults that "just work" and allow configuration, but I'm mildly leaning towards "this is a (almost) zero-downside default so its fine to not support configuration"

I will defer to whatever you guys want here. No strong opinion on my part.

src/refiners/fluxion/layers/attentions.py
brycedrennan: I made the comment just explain the default and moved it to `_sliced_attention`

src/refiners/fluxion/layers/attentions.py
limiteinductive: It is a matter of where it is happening in the abstraction layer. For instance, the `ScaledDotProductAttention` is the basis of other foundational models. When Mistral-7b is implemented, I'm unsure that slicing 4096 is necessarily a reasonable default. 

But it's completely ok to put it at the level of the SDUnet.

src/refiners/fluxion/layers/attentions.py
deltheil: For this [exact same reason](https://github.com/finegrain-ai/refiners/pull/130#discussion_r1412100335), I think you should either delete this comment or move it at a higher-level - UNet/SD - depending on what's decided to act on this setting

src/refiners/fluxion/layers/attentions.py
brycedrennan: I suspect its a reasonable default anywhere because its still a big chunk of memory so unlikely to have large performance impact but also takes the memory-pressure edge off of any large attention operations.  However I understand "my gut" is not the strongest argument.

src/refiners/fluxion/layers/attentions.py
brycedrennan: okay I will delete this comment even though I think it adds value :-) 
image

src/refiners/fluxion/layers/attentions.py
brycedrennan: Let me know. I'd be fine setting it to None just to get the PR merged.

brycedrennan:
Done. also yeah I have no idea how it interacts with grad

src/refiners/fluxion/layers/attentions.py
brycedrennan: comment removed

limiteinductive (approved):
lgtm

deltheil (approved):

PR #139, opened 2023-12-07
Feat: Prompt weighting 

GabrielVidal1:
Hi!

I wanted to start to work on the prompt weighting feature, having done something similar in the past, however the feature specs are really vague (just the words "Prompt weighting" in the bounties sheet) so I just wanted to clarify what was being expected.

My guess is the ability to provide multiple prompts (that can be weighted) to generate a single image. All the text embeddings and multiplied by a coefficient, and then added together.

For the syntax, I wanted to go for something like the [Midjourney multi prompting](https://docs.midjourney.com/docs/multi-prompts) syntax, so separating diffenrent sub-prompts by "::" or adding a coefficient after the "::" for the preceding prompt, I added some examples in the test_prompt_weighting.py file.

It can be done by just adding a function to parse the prompt, and return a list of tuple (prompt, weight) to the compute_clip_text_embedding of stable_diffusion_1 and stable_diffusion_xl


limiteinductive:
Hey! Thank you for locking the bounty. I like the Midjourney prompt weighting syntax, but we could do a poll on the Discord server to know what people prefer.

This needs to be a draft PR since there is no logic.

deltheil:
Hey @GabrielVidal1, just to confirm: are you still working on it? If not, that's fine - but then please close the PR :) Thanks!

GabrielVidal1:
Hey, I just pushed my current code, so you can see how things are done :)
Just didn't have the time to test it yet, I'm planning to in the following week or so

brycedrennan:
Nice, so if I'm reading this correctly you did tackle the more advanced weighting process where token weights are applied in context

PR #147, opened 2023-12-12
Change fl.Slicing Implementation

limiteinductive:
The current interface for fl.Slicing currently has two flaws:
  -> It follows Pytorch's  .narrow interface, which is confusing.
  -> It cannot perform slices with implicit lengths like this `x[1:]`.
  
  In this PR, I propose to make fl.Slicing follow more closely what we could expect from a slice method with `dim`, `start`, `end`, and `step` keywords.

### DIFF
--- a/src/refiners/fluxion/layers/basics.py
+++ b/src/refiners/fluxion/layers/basics.py
@@ -84,14 +84,31 @@ class Permute(Module):
 
 
 class Slicing(Module):
-    def __init__(self, dim: int, start: int, length: int) -> None:
+    def __init__(self, dim: int = 0, start: int = 0, end: int = -1, step: int = 1) -> None:
         super().__init__()
         self.dim = dim
         self.start = start
-        self.length = length
+        self.end = end
+        self.step = step
 
     def forward(self, x: Tensor) -> Tensor:
-        return x.narrow(self.dim, self.start, self.length)
+        dim_size = x.shape[self.dim]
+        start = self.start if self.start >= 0 else dim_size + self.start
+        end = self.end if self.end >= 0 else dim_size + self.end
+        start = max(min(start, dim_size), 0)
+        end = max(min(end, dim_size), 0)
+        if start >= end:
+            return self.get_empty_slice(x)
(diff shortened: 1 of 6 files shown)


src/refiners/fluxion/layers/basics.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, dim: int, start: int, length: int) -> None:
         self.length = length
 
     def forward(self, x: Tensor) -> Tensor:
+        if self.length < 0:
+            return x.narrow(self.dim, self.start, x.shape[self.dim] - self.start + self.length)
```
deltheil, line 95: Off-by-one error I guess?

```
>>> import refiners.fluxion.layers as fl
>>> import torch
>>> m = fl.Slicing(dim=2, start=2, length=-1)
>>> x = torch.tensor([[[0, 1, 2, 3, 4, 5, 6]]])
>>> y = m(x)
>>> z = x[:, :, 2:]
>>> y
tensor([[[2, 3, 4, 5]]])
>>> z
tensor([[[2, 3, 4, 5, 6]]])
```

deltheil (approved):

Laurent2916:
I think there is still an off-by-one error:
```python
>>> import torch
>>> import refiners.fluxion.layers as fl
>>> data = torch.randn(2, 1000, 400)
>>> slice = fl.Slicing(dim=1, start=1)
>>> slice(data).shape
torch.Size([2, 998, 400])
>>> data[:, 1:, :].shape
torch.Size([2, 999, 400])
```

limiteinductive:
> I think there is still an off-by-one error @limiteinductive:
> 
> ```python
> >>> import torch
> >>> import refiners.fluxion.layers as fl
> >>> data = torch.randn(2, 1000, 400)
> >>> slicing = fl.Slicing(dim=1, start=1)
> >>> slicing(data).shape
> torch.Size([2, 998, 400])
> >>> data[:, 1:, :].shape
> torch.Size([2, 999, 400])
> ```

Good catch! The default for the `end` param should be `None` and not `-1`. PR incoming

PR #152, opened 2023-12-14
refactor: convert prepare-test-weights.sh to python

brycedrennan:
This does feel a lot nicer to run but it is admittedly a mostly useless refactor (in that the previous version was mostly working).   

This is in preparation for wanting to re-use components of this for a similar script that outputs all the needed mapping to do load-time conversion. 

Ran successfully to completion. But on a repeat run `convert_unclip` didn't pass the hash check for some reason.

**Features**
- ~~fixes inpainting model download urls. some of the urls for the bash script were incorrect and the errors ignored~~
- skips downloading inpainting safety model and text encoder since we don't need to convert those
- shows a progress bar for downloads
- skips downloading existing files
- uses a temporary file to prevent partial downloads
- can do a dry run to check if url is valid `DRY_RUN=1 python scripts/prepare_test_weights.py`
- displays the downloaded file hash
- checks if converted weights exist, skips conversion if hash matches

Further refactoring could be done:
 - simplify the downloads to just a list of files and destinations (and possibly hashes)
 - use commit references instead of `main` branch names in urls for better reliability
 - put the converted weights in a separate folder

But I wanted to leave the script in a format that would still look somewhat recognizable.

**Screenshots**
image
image
image


scripts/conversion/convert_diffusers_autoencoder_kl.py
```diff
@@ -19,7 +19,7 @@ class Args(argparse.Namespace):
 def setup_converter(args: Args) -> ModelConverter:
     target = LatentDiffusionAutoencoder()
     source: nn.Module = AutoencoderKL.from_pretrained(  # type: ignore
-        pretrained_model_name_or_path=args.source_path, subfolder=args.subfolder
+        pretrained_model_name_or_path=args.source_path, subfolder=args.subfolder, low_cpu_mem_usage=False
```
brycedrennan, line 22: This just removes a lot of annoying messages from the console asking us to `pip install accelerate`

scripts/prepare_test_weights.py
```diff
@@ -0,0 +1,574 @@
...
+def human_readable_size(size: int | float, decimal_places: int = 2) -> str:
+    for unit in ["B", "KB", "MB", "GB", "TB", "PB"]:
+        if size < 1024.0:
+            break
+        size /= 1024.0
+    return f"{size:.{decimal_places}f}{unit}"  # type: ignore
+
+
+def download_sd_text_encoder(hf_repo_id: str = "runwayml/stable-diffusion-v1-5", subdir: str = "text_encoder"):
+    encoder_filename = "model.safetensors" if "inpainting" not in hf_repo_id else "model.fp16.safetensors"
```
brycedrennan, line 144: Corrects filename

scripts/prepare_test_weights.py
```diff
@@ -0,0 +1,574 @@
...
+    base_folder = os.path.join(test_weights_dir, hf_repo_id)
+
+    subdir = "feature_extractor"
+    download_file(
+        f"https://huggingface.co/{hf_repo_id}/raw/main/{subdir}/preprocessor_config.json",
+        os.path.join(base_folder, subdir),
+    )
+
+    subdir = "safety_checker"
+    safety_filename = "model.safetensors" if "inpainting" not in hf_repo_id else "model.fp16.safetensors"
```
brycedrennan, line 197: Corrects filename

scripts/prepare_test_weights.py
catwell: True (and below too) but maybe the best fix would be not to download it at all since we never convert it. We only need the unet from the inpainting model.

scripts/conversion/convert_diffusers_autoencoder_kl.py
catwell: Good idea, but maybe add a comment in the script to explain that.

brycedrennan: Interesting, I would have thought the auto formatter made that choice for me. TIL

scripts/prepare_test_weights.py
brycedrennan: image
Done

scripts/conversion/convert_diffusers_autoencoder_kl.py
brycedrennan: Done

scripts/prepare_test_weights.py
```diff
@@ -0,0 +1,577 @@
...
+DRY_RUN=1 python scripts/prepare_test_weights.py
+"""
+import hashlib
+import os
+import subprocess
+import sys
+from urllib.parse import urlparse
+
+import requests
+from tqdm import tqdm
```
catwell, line 14: Since you import those directly you should add them explicitly as dependencies in the `conversion` extra.

catwell (changes requested):
> check if converted weights exist and match hash. skip if thats the case

This one would be useful (in combination with skipping the downloads if files exists, it will make re-running the script very cheap). I suggest you add it.

I tried the script and conversion failed for SAM (I did not yet investigate if this is due to the script or not):

Screenshot from 2023-12-14 18-30-04

You will also need to change `CONTRIBUTING.md` which references the old script.

scripts/prepare_test_weights.py
brycedrennan: done

brycedrennan:
> check if converted weights exist and match hash. skip if thats the case
Done.

catwell (approved):
There are still things that could be improved, for instance: why shell out to curl in `convert_preprocessors` instead of using `download_file`?

However in its current state it works and it is undeniably an improvement, so let's merge and change it later if needed!

PR #166, opened 2024-01-08
fix CI on push

catwell:
for reference, documentation is here:
ttps://docs.github.com/en/actions/learn-github-actions/contexts#github-context


deltheil (approved):
Thanks!!

PR #167, opened 2024-01-08
feature: Color Palette Adapter

Laurent2916:
Hey!

This PR implements a `PaletteAdapter` (only tested on `SD1UNet` for now).
Schematically it looks a bit like this:
image
The `ColorEncoder` is basically how I described it in the gdoc, and the `CrossAttentionAdapter` is inspired from the ones found in the existing `IPAdapter`.

I implemented an `AdapterLatentDiffusionTrainer`, which is almost complete (I have a couple questions on image pre-processing, and how to train on each palette "size"). I tested it on my CPU and it should be good to go.

The trainer uses a `PaletteDataset` which is basically my huggingface dataset ([1aurent/unsplash-lite-palette](https://huggingface.co/datasets/1aurent/unsplash-lite-palette)) with a couple `map`.

I'll also need some computing, since I'm part of the "GPU poor" gang.


src/refiners/foundationals/latent_diffusion/palette_adapter.py
```diff
@@ -0,0 +1,314 @@
...
+        self.max_colors = max_colors
+
+        super().__init__(
+            fl.UseContext("palette", "colors"),
+            # (n_colors, 3): uint8
+            fl.Concatenate(
+                fl.Chain(
+                    SinusoidalEmbedding(dim_embedding=dim_sinusoids),
+                    # (n_colors, 3, dim_sinusoids): float32
+                    fl.Reshape(-1, dim_sinusoids * 3),
```
limiteinductive, line 177: is that reshape correct

src/refiners/foundationals/latent_diffusion/palette_adapter.py
```diff
@@ -0,0 +1,314 @@
...
+                    device=device,
+                    dtype=dtype,
+                ),  # (1, dim_embeddings)
+                dim=1,
+            ),
+            # (n_colors + 1, dim_embeddings)
+            FillZeroRightPad(
+                max_length=max_colors + 1,
+                dim=1,
+            ),
```
limiteinductive, line 199:
```suggestion
                    fl.Reshape(dim_sinusoids * 3),
                    # (n_colors, 3 * dim_sinusoids)
                    FeedForward(
                        input_dim=dim_sinusoids * 3,
                        intermediate_dim=dim_sinusoids * 3,
                        output_dim=dim_embeddings,
                        device=device,
                        dtype=dtype,
                    ),
                    # (n_colors, dim_embeddings)
                ),
                EOSToken(
                    embedding_dim=dim_embeddings,
                    device=device,
                    dtype=dtype,
                ),  # (1, dim_embeddings)
                dim=0,
            ),
            # (n_colors + 1, dim_embeddings)
            FillZeroRightPad(
                max_length=max_colors + 1,
                dim=0,
            ),
```

src/refiners/foundationals/latent_diffusion/palette_adapter.py
```diff
@@ -0,0 +1,314 @@
...
+        self,
+        embedding_dim: int,
+        device: torch.device | str | None = None,
+        dtype: torch.dtype | None = None,
+    ) -> None:
+        self.embedding_dim = embedding_dim
+
+        super().__init__(
+            fl.Parameter(
+                *(1, embedding_dim),
```
limiteinductive, line 54:
```suggestion
                embedding_dim,
```

scripts/training/finetune-ldm-palette-adapter.py
```diff
@@ -0,0 +1,412 @@
...
+from refiners.training_utils.config import BaseConfig
+from refiners.training_utils.latent_diffusion import (
+    LatentDiffusionConfig,
+    TestDiffusionConfig,
+    resize_image,
+    sample_noise,
+)
+from refiners.training_utils.trainer import Trainer
+
+Image.MAX_IMAGE_PIXELS = None
```
deltheil, line 32: Adding a comment would not hurt (about why you turn off this "decompression bombs" safety feature - is it really needed?)

scripts/training/finetune-ldm-palette-adapter.py
```diff
@@ -0,0 +1,412 @@
...
+    """
+
+    def __init__(self, trainer: "AdapterLatentDiffusionTrainer") -> None:
+        super().__init__()
+        self.trainer = trainer
+        self.dataset = self.load_huggingface_dataset()
+
+    @staticmethod
+    def download_image(
+        data: dict[str, list[Any]],  # "batched"
```
deltheil, line 100: Or `download_images` since it's batched (same applies below)

scripts/training/finetune-ldm-palette-adapter.py
```diff
@@ -0,0 +1,412 @@
...
+        dataset = datasets.load_dataset(  # type: ignore
+            path=dataset_config.hf_repo,
+            revision=dataset_config.revision,
+            split=dataset_config.split,
+        )
+        dataset = dataset.select(list(range(100)))  # type: ignore # FIXME: temporary
+
+        # download images from urls
+        dl_manager = datasets.DownloadManager()
+        dataset = dataset.map(  # type: ignore
```
deltheil, line 153: How long does this take at cold start? (rough number)

scripts/training/finetune-ldm-palette-adapter.py
```diff
@@ -0,0 +1,412 @@
...
+class AdapterLatentDiffusionConfig(BaseConfig):
+    """Finetunning configuration.
+
+    Contains the configs of the dataset, the latent diffusion model and the adapter.
+    """
+
+    dataset: DatasetConfig
+    ldm: LatentDiffusionConfig
+    adapter: AdapterConfig
+    test_ldm: TestPaletteDiffusionConfig
```
deltheil, line 76: Still not used/TODO?

scripts/training/finetune-ldm-palette-adapter.py
Laurent2916: I'll add a comment, the default limit for [MAX_IMAGE_PIXELS](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.MAX_IMAGE_PIXELS) (89_478_485 pixels) is often lower than some of the images of the unsplash-lite dataset. For example [this image](https://images.unsplash.com/photo-1572371733185-448e1d92ea56) is 16384*10923=178_962_432 pixels.

scripts/training/finetune-ldm-palette-adapter.py
deltheil: Also, how does DownloadManager handles download failures? Is there a retry mechanism or something?

scripts/training/finetune-ldm-palette-adapter.py
Laurent2916: In my case, downloading the 24998 images of the [1aurent/unsplash-lite-palette](https://huggingface.co/datasets/1aurent/unsplash-lite-palette) dataset took about 20-30 minutes, though this obviously depends of your internet speed. Also, the images are a bunch of small files (about 2 to 5 MB each), so this doesn't help.
If not a cold start, this operation is instantaneous since the result of [`datasets.map`](https://huggingface.co/docs/datasets/main/en/package_reference/main_classes#datasets.Dataset.map) operations are cached.

src/refiners/foundationals/latent_diffusion/palette_adapter.py
```diff
@@ -0,0 +1,367 @@
...
+                    SinusoidalEmbedding(
+                        dim_embedding=dim_sinusoids,
+                        dtype=dtype,
+                    ),
+                    # (batch, n_colors, 3, dim_sinusoids): float32
+                    fl.Reshape(-1, dim_sinusoids * 3),
+                    # (batch, n_colors, 3 * dim_sinusoids)
+                    FeedForward(
+                        input_dim=dim_sinusoids * 3,
+                        intermediate_dim=dim_sinusoids * 3,
```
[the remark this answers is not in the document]
Laurent2916: No reasons, I honestly don't know what to pick for `intermediate_dim`, so I made it equal to `input_dim`.

scripts/training/finetune-ldm-palette-adapter.py
Laurent2916: I'll set `MAX_IMAGE_PIXELS` to 200_000_000:
```python
>>> ds = datasets.load_dataset("1aurent/unsplash-lite", split="train")
>>> ds2 = ds.map(lambda x: {"width": x["photo"]["width"], "height": x["photo"]["height"]}, remove_columns=ds.column_names)
>>> ds3 = ds2.map(lambda x: {"pixels": x["width"] * x["height"]})
>>> max(ds3["pixels"])
178962432

File: scripts/training/finetune-ldm-palette-adapter.py (hunk 6 above)
Comment (deltheil) [reply scripts/training/finetune-ldm-palette-adapter.py:154]: Good to know, thanks. Re: cache: true. Perhaps you want to expose the cache info settings (via the config) to choose a different path at some point.

Have you considered doing this offline once and for all and store a new dataset with image bytes bundled? Pros/cons?

File: src/refiners/foundationals/latent_diffusion/palette_adapter.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (deltheil) [reply src/refiners/foundationals/latent_diffusion/palette_adapter.py:185]: I am asking because with e.g. ViT it would typically be a factor=4 cc @limiteinductive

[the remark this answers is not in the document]
Comment (limiteinductive) [reply src/refiners/foundationals/latent_diffusion/palette_adapter.py:185]: pretty hard to come up with an heuristic here, I don't think we would need that much dims, but better safe than sorry (and it doesn't cost that much on a single layer)

File: scripts/training/finetune-ldm-palette-adapter.py (hunk 6 above)
Comment (Laurent2916) [reply scripts/training/finetune-ldm-palette-adapter.py:154]: > Perhaps you want to expose the cache info settings (via the config) to choose a different path at some point.

The cache's location can be selected by modifying some environment variables, like [`HF_HOME`](https://huggingface.co/docs/huggingface_hub/main/en/package_reference/environment_variables#hfhome) or [`HF_DATASETS_CACHE`](https://huggingface.co/docs/datasets/cache#cache-directory).

> Have you considered doing this offline once and for all and store a new dataset with image bytes bundled? Pros/cons?

On advantage would be downloading the dataset in big chunks, thus potentially making the download faster.
If the images are bundled in the dataset, this would mean that images are loaded from the big `.arrow` files, and I don't know if this is faster than loading them from their respective `.png`/`.jpeg` files.
Though this doesn't matter much since after the `resize_images` mapping, the images are bundled inside `.arrow` files anyway. So this would only impact cold start.

Also the dataset's license doesn't allow redistributing the images, so I couldn't share it (publicly) on the huggingface hub.
> The Unsplash Dataset is made available for research purposes. [It cannot be used to redistribute the images contained within](https://github.com/unsplash/datasets/blob/master/TERMS.md).

File: src/refiners/foundationals/latent_diffusion/palette_adapter.py [hunk 9]
```diff
@@ -0,0 +1,367 @@
...
+        self.dim = dim
+        super().__init__()
+
+    def forward(
+        self,
+        x: torch.Tensor,
+    ) -> torch.Tensor:
+        pad = [0] * 2 * x.ndim
+        dim_index = 1 + 2 * (x.ndim - 1) - 2 * self.dim
+        pad[dim_index] = self.max_length - x.size(self.dim)
```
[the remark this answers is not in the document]
Laurent2916: Well we need all the palettes to be of same size to batch them. If one of them exceeds `max_length` we could either raise an exception of "crop" it to `max_length`.
I guess we `assert` this when the user sets the context.

scripts/training/finetune-ldm-palette-adapter.py
Laurent2916: [`DownloadManager.download()`](https://huggingface.co/docs/datasets/v2.16.1/en/package_reference/builder_classes#datasets.DownloadManager.download) calls at one point [`cached_path`](https://github.com/huggingface/datasets/blob/7b2bcd76457de720454c3ac304f2ed5c6f40acaa/src/datasets/utils/file_utils.py#L151) which raises ConnectionErrors. It looks like a `max_retries` (defaults to 1) (and a bunch of other options) can be configured through a [`DownloadConfig`](https://huggingface.co/docs/datasets/v2.16.1/en/package_reference/builder_classes#datasets.DownloadConfig).

scripts/training/finetune-ldm-palette-adapter.py
Laurent2916: fixed by 2abd185

src/refiners/foundationals/latent_diffusion/palette_adapter.py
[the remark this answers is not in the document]
Laurent2916: fixed in recent commits, the palettes are now batched using [`torch.nn.utils.rnn.pad_sequence`](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_sequence.html)

PR #168, opened 2024-01-08
Refactor CrossAttentionAdapter to work with context.

limiteinductive:
Currently, CrossAttentionAdapter works by slicing the condition text_embedding to get the image_embedding that has been stacked to it. This can interfere with other adapters like `ControlNet,` for instance, since it modifies its text condition, too.

Weights will differ, but old weights should stay compatible with this implementation.

The API for using IpAdapter is simplified like this:
```diff
clip_image_embedding = ip_adapter.compute_clip_image_embedding(ip_adapter.preprocess_image(image_prompt))

-  negative_text_embedding, conditional_text_embedding = clip_text_embedding.chunk(2)
- negative_image_embedding, conditional_image_embedding = clip_image_embedding.chunk(2)
-
-  clip_text_embedding = torch.cat(
-     (
-          torch.cat([negative_text_embedding, negative_image_embedding], dim=1),
-          torch.cat([conditional_text_embedding, conditional_image_embedding], dim=1),
-      )
-  )
+  ip_adapter.set_clip_image_embedding(clip_image_embedding)
```

You can now also directly set the scale as an attribute:
```python
ip_adapter.scale = 0.7
```


scripts/conversion/convert_segment_anything.py
deltheil, line 1: Should not be impacted: are you up-to-date with main?

deltheil (changes requested):
Please see comment re: SAM weights conversion

deltheil (approved):

PR #171, opened 2024-01-09
Make summarize_tensor robust to non-float dtypes 

piercus:
Current implementation of summarize_tensor can raise secondary error while obfuscating the original error.

## Actual

When doing some experiments, the stack was displaying a secondary error

```
Traceback (most recent call last):
  File "/home/pierre/dev/refiners/scripts/training/finetune-ldm-color-palette.py", line 15, in <module>
    trainer.train()
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 89, in inner_wrapper
    result = func(*args, **kwargs)
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 528, in train
    self.epoch()
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 512, in epoch
    self.step(batch=batch)
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 503, in step
    loss = self.compute_loss(batch=batch)
  File "/home/pierre/dev/refiners/src/refiners/training_utils/color_palette.py", line 234, in compute_loss
    prediction = self.unet(noisy_latents)
  File "/home/pierre/anaconda3/envs/refiners/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/pierre/anaconda3/envs/refiners/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/pierre/dev/refiners/src/refiners/fluxion/layers/chain.py", line 279, in forward
    result = self._call_layer(layer, name, *intermediate_args)
  File "/home/pierre/dev/refiners/src/refiners/fluxion/layers/chain.py", line 273, in _call_layer
    raise ChainError(message) from None
refiners.fluxion.layers.chain.ChainError: RuntimeError:
   File "/home/pierre/dev/refiners/src/refiners/fluxion/utils.py", line 199, in summarize_tensor
    f"mean={tensor.mean():.2f}",

mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long
```

## Expected

I'm expecting to get to orignal error in the stack

Which in my case was 

`Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)`

## Solution

This PR implements a fix for this

### DIFF
--- a/src/refiners/fluxion/utils.py
+++ b/src/refiners/fluxion/utils.py
@@ -195,10 +195,9 @@ def summarize_tensor(tensor: torch.Tensor, /) -> str:
         f"max={tensor.max():.2f}",  # type: ignore
     ]
 
-    # Only add mean, std, and norm for float and complex tensors
-    if tensor.dtype == torch.float or tensor.dtype == torch.double:
-        info_list.extend([f"mean={tensor.mean():.2f}", f"std={tensor.std():.2f}", f"norm={norm(x=tensor):.2f}"])
+    if tensor.dtype == torch.float or tensor.dtype == torch.complex:
+        info_list.append(f"mean={tensor.mean():.2f}")
 
-    info_list.append(f"grad={tensor.requires_grad}")
+    info_list.extend([f"std={tensor.std():.2f}", f"norm={norm(x=tensor):.2f}", f"grad={tensor.requires_grad}"])
 
     return "Tensor(" + ", ".join(info_list) + ")"
--- a/src/refiners/fluxion/utils.py
+++ b/src/refiners/fluxion/utils.py
@@ -187,20 +187,18 @@ def save_to_safetensors(path: Path | str, tensors: dict[str, Tensor], metadata: 
 
 
 def summarize_tensor(tensor: torch.Tensor, /) -> str:
-    return (
-        "Tensor("
-        + ", ".join(
-            [
-                f"shape=({', '.join(map(str, tensor.shape))})",
-                f"dtype={str(object=tensor.dtype).removeprefix('torch.')}",
-                f"device={tensor.device}",
-                f"min={tensor.min():.2f}",  # type: ignore
-                f"max={tensor.max():.2f}",  # type: ignore
-                f"mean={tensor.mean():.2f}",
-                f"std={tensor.std():.2f}",
-                f"norm={norm(x=tensor):.2f}",
-                f"grad={tensor.requires_grad}",
-            ]
-        )
-        + ")"
-    )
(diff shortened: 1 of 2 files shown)


limiteinductive (dismissed):
Seems good to me, important fix

src/refiners/fluxion/utils.py
```diff
@@ -187,20 +187,18 @@ def save_to_safetensors(path: Path | str, tensors: dict[str, Tensor], metadata:
...
-    )
+    info_list = [
+        f"shape=({', '.join(map(str, tensor.shape))})",
+        f"dtype={str(object=tensor.dtype).removeprefix('torch.')}",
+        f"device={tensor.device}",
+        f"min={tensor.min():.2f}",  # type: ignore
+        f"max={tensor.max():.2f}",  # type: ignore
+    ]
+
+    # Only add mean, std, and norm for float and complex tensors
```
limiteinductive, line 198: I don't think this comment is really helpful (and is misleading since you didn't add complex types to the condition?)

piercus:
Thanks for your feedback @limiteinductive.

I feel it worth unit testing summarize_tensor.
It will prevent any regression here (which might not be visible elsewhere in the CI).

src/refiners/fluxion/utils.py
```diff
@@ -187,20 +187,22 @@ def save_to_safetensors(path: Path | str, tensors: dict[str, Tensor], metadata:
...
-                f"mean={tensor.mean():.2f}",
-                f"std={tensor.std():.2f}",
-                f"norm={norm(x=tensor):.2f}",
-                f"grad={tensor.requires_grad}",
             ]
         )
-        + ")"
-    )
+
+    if tensor.dtype == torch.float or tensor.is_complex():
```
limiteinductive, line 203: Now that I think of it, this seems too restrictive: this will filter out torch.float16 tensors, and many helpful other types; furthermore, it makes sense to compute the mean, std, and norm of an int tensor, for instance. So, is the best solution to cast the tensor to float to compute those values? I let you come up with the best solution.

limiteinductive (changes requested):

limiteinductive (approved):

deltheil:
Thanks @piercus!

PR #165, opened 2024-01-06
Train a Color Palette adapter

piercus:
Hello Refiners,

Please find my first commits on Color Palette Adapter training.
It's still work in progress.

## Current status

I've tried to set up the main components following guidelines from https://docs.google.com/document/d/1rhRr_1gscyW_qK5558z4lnrehNC216cnGen_79gh77s/edit#heading=h.d7ff3urp6olq

I made some comments in the code about points I'd like to discuss.

## Blocking point

When running `python scripts/training/finetune-ldm-color-palette.py configs/finetune-color-palette.toml` i'm blocked by my GPU memory `OutOfMemoryError`.

```bash
Traceback (most recent call last):
  File "/home/pierre/dev/refiners/scripts/training/finetune-ldm-color-palette.py", line 313, in <module>
    trainer.train()
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 89, in inner_wrapper
    result = func(*args, **kwargs)
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 528, in train
    self.epoch()
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 512, in epoch
    self.step(batch=batch)
  File "/home/pierre/dev/refiners/src/refiners/training_utils/trainer.py", line 503, in step
    loss = self.compute_loss(batch=batch)
  File "/home/pierre/dev/refiners/scripts/training/finetune-ldm-color-palette.py", line 223, in compute_loss
    prediction = self.unet(noisy_latents)
  File "/home/pierre/anaconda3/envs/refiners/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/pierre/anaconda3/envs/refiners/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/pierre/dev/refiners/src/refiners/fluxion/layers/chain.py", line 279, in forward
    result = self._call_layer(layer, name, *intermediate_args)
  File "/home/pierre/dev/refiners/src/refiners/fluxion/layers/chain.py", line 273, in _call_layer
    raise ChainError(message) from None
refiners.fluxion.layers.chain.ChainError: 

OutOfMemoryError:
[...]

PR #173, opened 2024-01-11
display module dtype and device

piercus:
## Context

In the context of [#165](https://github.com/finegrain-ai/refiners/pull/165#issuecomment-1882887690), I'm splitting the training on different GPUs.

While doing this, I faced a lot of errors like `Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!`.

## Actual

The stack trace gives information about input tensor device allocation, but not any information about the module's device & dtype information.

## Expected

Print the device and the dtype of modules in the stack trace.

### DIFF
--- a/src/refiners/fluxion/layers/module.py
+++ b/src/refiners/fluxion/layers/module.py
@@ -164,6 +164,10 @@ class WeightedModule(Module):
     def dtype(self) -> DType:
         return self.weight.dtype
 
+    def __repr__(self) -> str:
+        str = super().__repr__()
+        return f"{str} [device={self.device}, dtype={self.dtype}]"
+
 
 class TreeNode(TypedDict):
     value: str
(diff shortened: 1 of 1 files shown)


limiteinductive:
This clutters a lot the tree view.
image

I have two ideas to improve this matter:
 - show dtype/device only for `WeightedModule`
 - show only on the higher level `Chain` and hide this information except where it's different

limiteinductive (changes requested):
The idea is sound, let's find a better way to do it.

piercus:
This version is much more KISS and should be less verbose also.
Since `Chain` has `_show_only_tag()`. I feel we should show only tag for Chain.

src/refiners/fluxion/layers/module.py
```diff
@@ -164,6 +164,10 @@ def device(self) -> Device:
     def dtype(self) -> DType:
         return self.weight.dtype
 
+    def __repr__(self) -> str:
+        str = super().__repr__()
+        return f"{str} [device={self.device}, dtype={self.dtype}]"
+
```
limiteinductive, line 170: For it to show up in the error tree, you need to modify `__str__` 
```suggestion
    def __str__(self) -> str:
        return f"{super().__str__().removesuffix(')')}, device={self.device}, dtype={str(self.dtype).removeprefix('torch.')})"

```

limiteinductive (changes requested):

piercus:
Thanks for your help @limiteinductive

limiteinductive (approved):

PR #179, opened 2024-01-12
Use public url dataset in the configs/finetune-lora.toml

piercus:
## Context

Related to https://github.com/finegrain-ai/refiners/pull/165#discussion_r1445967855
I'm trying to PR, little by little, my progress on #165

## Actual

The hf repo in the main branch in [config/finetune-lora.toml](https://github.com/finegrain-ai/refiners/blob/main/configs/finetune-lora.toml#L52C1-L54C18) are private repository.

It's not plug and play for a newcomer on the lib.

## Expected

The lib examples should work using public hf repository.

## Solution

I suggest to use https://huggingface.co/datasets/1aurent/unsplash-lite-palette for now, but we also need to make the latent_diffusion Trainer compatible with url datasets as discussed in  https://github.com/finegrain-ai/refiners/pull/165#discussion_r1445967855

NB: The current code is not optimized for pre-downloading or batch-downloading. 
The main purpose here is to make it work and keep it simple.

### DIFF
--- a/configs/finetune-lora.toml
+++ b/configs/finetune-lora.toml
@@ -50,8 +50,9 @@ dropout_probability = 0.2
 use_gyro_dropout = false
 
 [dataset]
-hf_repo = "acme/images"
+hf_repo = "1aurent/unsplash-lite-palette"
 revision = "main"
+caption_key = "ai_description"
 
 [checkpointing]
 # save_folder = "/path/to/ckpts"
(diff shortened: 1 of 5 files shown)


limiteinductive:
I think this PR is obsolete with the new training_utils? wdyt @piercus

piercus:
@limiteinductive agreed

PR #181, opened 2024-01-14
Write basic unit tests for Trainer

limiteinductive:


tests/fluxion/layers/test_chain.py
deltheil, line 1: Your branch seems out of date with main

limiteinductive: should be fixed

deltheil (approved):

PR #193, opened 2024-01-19
fix: dpm_solver shouldn't be required to "start" on step 0

brycedrennan:
Example: img2img pipelines often start somewhere in the middle based on how much of the source image they want to keep.

It wouldn't surprise me that you guys know a better way to allow starting in the "middle" of a diffusion process, but this did get my test suite passing.

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/schedulers/dpm_solver.py
+++ b/src/refiners/foundationals/latent_diffusion/schedulers/dpm_solver.py
@@ -38,6 +38,7 @@ class DPMSolver(Scheduler):
         )
         self.estimated_data = deque([tensor([])] * 2, maxlen=2)
         self.last_step_first_order = last_step_first_order
+        self._first_step_has_been_run = False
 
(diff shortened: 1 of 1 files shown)


catwell:
Ah thanks, I missed that in [my previous PR](https://github.com/finegrain-ai/refiners/pull/191) indeed.

We need a test for this in refiners, ideally one that runs on CPU.

catwell:
Here is an alternative idea that does not have the problem of keeping a hidden state:

- we add a set_first_step on LatentDiffusionModel (or an extra parameter to set_num_inference_steps);
- we take it into account for init_latents and pass it to schedulers as an argument.

brycedrennan:
> Isn't doing it like this (with a scheduler attribute, like it was the case earlier) also a problem if you run several loops though?

You mean multiple image generations right?  I make a new scheduler instance for each generation.  In my mind, the schedulers already store state about the image generation via "num_inference_steps" (which can vary from generation to generation).

brycedrennan:
Looks good!

PR #194, opened 2024-01-19
training dependencies are test dependencies

catwell:


CONTRIBUTING.md
deltheil, line 1: Ideally we should have a way to express test dependencies as ~ `test = [ ... ] + training` but it looks like there is no direct way to do that in TOML. Let me know if you think about an alternative?

deltheil (approved):
LGTM (ideal solution would be to improve `[project.optional-dependencies]` - we can keep it for later)

CONTRIBUTING.md
catwell: I don't. For tests it makes sense to install everything anyway since we want everything to be covered.

PR #200, opened 2024-01-19
fix: summarize_tensor(tensor) when tensor.numel() == 0

piercus:
Fix following error
```
RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.
```


limiteinductive (approved):

deltheil:
@piercus thanks!

PR #201, opened 2024-01-19
make the first diffusion step a first class property of LDM & Schedulers

catwell:
Notes:

This changes the interface of SD in a backward-incompatible (but IMO simpler) way, as you can see from the changes in the README:

Screenshot from 2024-01-19 16-48-55

This also fixes #193.

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/schedulers/scheduler.py
+++ b/src/refiners/foundationals/latent_diffusion/schedulers/scheduler.py
@@ -33,6 +33,7 @@ class Scheduler(ABC):
         initial_diffusion_rate: float = 8.5e-4,
         final_diffusion_rate: float = 1.2e-2,
         noise_schedule: NoiseSchedule = NoiseSchedule.QUADRATIC,
+        first_inference_step: int = 0,
         device: Device | str = "cpu",
         dtype: DType = float32,
     ):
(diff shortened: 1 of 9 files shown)


deltheil (dismissed):

src/refiners/foundationals/latent_diffusion/schedulers/scheduler.py
```diff
@@ -68,9 +70,13 @@ def _generate_timesteps(self) -> Tensor:
         ...
 
     @property
-    def steps(self) -> list[int]:
+    def all_steps(self) -> list[int]:
         return list(range(self.num_inference_steps))
 
+    @property
+    def diffusion_steps(self) -> list[int]:
```
limiteinductive, line 77: This should be called "denoising_steps" since it corresponds to the reverse diffusion process.

src/refiners/foundationals/latent_diffusion/schedulers/scheduler.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         initial_diffusion_rate: float = 8.5e-4,
         final_diffusion_rate: float = 1.2e-2,
         noise_schedule: NoiseSchedule = NoiseSchedule.QUADRATIC,
+        first_diffusion_step: int = 0,
```
limiteinductive, line 36: ditto

src/refiners/foundationals/latent_diffusion/schedulers/scheduler.py
catwell: Renamed it to "inference_steps" for coherence with "num_inference_steps".

brycedrennan (approved):
This is great. 

I must confess I am tempted to try and make the `__call__` method in the scheduler stateless but that's a large paradigm shift with questionable value.  
```
    def __call__(
        self,
        x: Tensor,
        noise: Tensor,
        step: int,
        num_steps: int = 50,
        first_step: int = 0,
    ) -> Tensor:
```

deltheil (approved):

PR #203, opened 2024-01-20
refactor: rename noise => predicted_noise

brycedrennan:
noise => predicted_noise
denoised_x => renoised_x (denoised seems inaccurate since it does have noise added)

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/model.py
+++ b/src/refiners/foundationals/latent_diffusion/model.py
@@ -91,15 +91,17 @@ class LatentDiffusionModel(fl.Module, ABC):
         unconditional_prediction, conditional_prediction = self.unet(latents).chunk(2)
 
         # classifier-free guidance
-        noise = unconditional_prediction + condition_scale * (conditional_prediction - unconditional_prediction)
+        predicted_noise = unconditional_prediction + condition_scale * (
+            conditional_prediction - unconditional_prediction
+        )
         x = x.narrow(dim=1, start=0, length=4)  # support > 4 channels for inpainting
 
         if self.has_self_attention_guidance():
-            noise += self.compute_self_attention_guidance(
+            predicted_noise += self.compute_self_attention_guidance(
                 x=x, noise=unconditional_prediction, step=step, clip_text_embedding=clip_text_embedding, **kwargs
             )
 
-        return self.scheduler(x, noise=noise, step=step)
+        return self.scheduler(x, predicted_noise=predicted_noise, step=step)
 
     def structural_copy(self: TLatentDiffusionModel) -> TLatentDiffusionModel:
         return self.__class__(
--- a/src/refiners/foundationals/latent_diffusion/schedulers/euler.py
+++ b/src/refiners/foundationals/latent_diffusion/schedulers/euler.py
@@ -58,7 +58,7 @@ class EulerScheduler(Scheduler):
     def __call__(
         self,
         x: Tensor,
-        noise: Tensor,
+        predicted_noise: Tensor,
         step: int,
         generator: Generator | None = None,
         s_churn: float = 0.0,
@@ -72,13 +72,15 @@ class EulerScheduler(Scheduler):
 
         gamma = min(s_churn / (len(self.sigmas) - 1), 2**0.5 - 1) if s_tmin <= sigma <= s_tmax else 0
 
-        alt_noise = torch.randn(noise.shape, generator=generator, device=noise.device, dtype=noise.dtype)
-        eps = alt_noise * s_noise
+        noise = torch.randn(
+            predicted_noise.shape, generator=generator, device=predicted_noise.device, dtype=predicted_noise.dtype
+        )
+        eps = noise * s_noise
         sigma_hat = sigma * (gamma + 1)
         if gamma > 0:
             x = x + eps * (sigma_hat**2 - sigma**2) ** 0.5
 
-        predicted_x = x - sigma_hat * noise
+        predicted_x = x - sigma_hat * predicted_noise
 
         # 1st order Euler
         derivative = (x - predicted_x) / sigma_hat
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,5 +96,5 @@ black = true
 [tool.pyright]
 include = ["src/refiners", "tests", "scripts"]
 strict = ["*"]
-exclude = ["**/__pycache__"]
+exclude = ["**/__pycache__", "tests/weights"]
 reportMissingTypeStubs = "warning"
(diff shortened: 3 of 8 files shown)


limiteinductive:
> noise => predicted_noise denoised_x => renoised_x (denoised seems inaccurate since it does have noise added)

A propos `renoised_x,` I get that since it is not fully denoised but only slightly denoised, `denoised_x` is a bit ambiguous; maybe `intermediary_denoised_x` or `partially_denoised_x` is more accurate, but they feel a bit too much for me.

For me, `renoising` suggests that we're stationary in the level of noise

limiteinductive:
> noise => predicted_noise denoised_x => renoised_x (denoised seems inaccurate since it does have noise added)

and noise => predicted_noise  is ok for me

brycedrennan:
Yeah that makes sense. I think just calling it `next_x` would be less misleading.   Of all the ideas so far though, I do like `slightly_denoised_x` best- it conveys exactly what it is, even if its a long name. It is at least not used outside the function.

I was suspicious that labeling it "denoised" contributed to the bug of it being returned in the final step.  I'm also suspicious that a similar bug exists in the other schedulers but I haven't investigated enough yet.

These changes were in the context of an not-finished pull request where I make the following change in ddim (as was looking to see if similar changes applied in the other schedulers):

```python
        if step == self.num_inference_steps - 1:
            noise_factor = 0
 ```
 to 
```python        
        if step == self.num_inference_steps - 1:
            return predicted_x
```

catwell:
AFAIK the additional noise term is not really a "bug", at least not in refiners, it's a bad decision in the original implementation of SD (CompVis) which we used as a reference. See [the documentation of the `set_alpha_to_one` option in Diffusers](https://github.com/huggingface/diffusers/blob/c7df846decbe9aa335a5292df9e5c87c5fb56d2d/src/diffusers/schedulers/scheduling_ddim.py#L155-L162), for instance. It did not come from the naming.

Regarding other schedulers, the [ability to use a 1st order update for the last step in DPM solver that I added last week is somehow similar](https://github.com/finegrain-ai/refiners/pull/191). It will remove the same kind of noise which is very visible especially if you use SDXL with few timesteps. For now we decided against making it the default.

src/refiners/foundationals/latent_diffusion/schedulers/euler.py
```diff
@@ -72,13 +72,15 @@ def __call__(
 
         gamma = min(s_churn / (len(self.sigmas) - 1), 2**0.5 - 1) if s_tmin <= sigma <= s_tmax else 0
 
-        alt_noise = torch.randn(noise.shape, generator=generator, device=noise.device, dtype=noise.dtype)
+        alt_noise = torch.randn(
```
deltheil, line 75: Now that `noise` is (properly) renamed into `predicted_noise` I would be tempted to use `noise` for this one (stochastic sampler), i.e. the `alt_` prefix is not needed anymore. WDYT?

brycedrennan:
I have reduced the scope of this pull request to just `noise` => `predicted_noise`

brycedrennan:
@catwell thanks for the context - that makes sense

src/refiners/foundationals/latent_diffusion/schedulers/euler.py
deltheil: cc @brycedrennan

brycedrennan: Sounds good to me. 

If I'm reading the code correctly, with default settings, the noise is never used anyways.

Maybe never used at all. https://github.com/huggingface/diffusers/issues/4785#issuecomment-1694233294

brycedrennan: I've surely gone mad.  Does the whole thing simplify down to:

```python
    def __call__(
        self,
        x: Tensor,
        predicted_noise: Tensor,
        step: int,
        generator: Generator | None = None,
    ) -> Tensor:
        assert self.first_inference_step <= step < self.num_inference_steps, "invalid step {step}"

        delta_t = self.sigmas[step + 1] - self.sigmas[step]
        return x + predicted_noise * delta_t
 ```

Comment (catwell) [reply src/refiners/foundationals/latent_diffusion/schedulers/euler.py:75]: Disclaimer: I have basically not looked into the Euler scheduler yet (it was a contribution). But given the name it makes sense that it simplifies to this (Euler's method) when you remove the stochastic part (churn = 0).

There is [an in depth explanation about this scheduler](https://isamu-website.medium.com/understanding-k-diffusion-from-their-research-paper-and-source-code-55ae4aa802f) on @isamu-isozaki's blog by the way!

Comment (brycedrennan) [reply src/refiners/foundationals/latent_diffusion/schedulers/euler.py:75]: Yes very in depth!

**Comment by brycedrennan** [comment]
We've discussed a few changes here and I'm happy to do all or none of them.  Let me know what is desired to get this closed out.

**Comment by deltheil** [comment]
> We've discussed a few changes here and I'm happy to do all or none of them. Let me know what is desired to get this closed out.

How about this:

1. Let's keep the `noise` => `predicted_noise` renaming (makes things more readable)
2. Re: `schedulers/euler.py` => let's keep the implementation as-is, except the `noise` => `predicted_noise` and `alt_noise` => `noise` renamings. Rationale: we're going to revamp it (completely) at some point

@brycedrennan does that work for you? Many thanks!

File: src/refiners/foundationals/latent_diffusion/schedulers/euler.py (hunk 1 above)
Comment (isamu-isozaki) [reply src/refiners/foundationals/latent_diffusion/schedulers/euler.py:75]: @catwell Thanks for the exposure!

**Comment by brycedrennan** [comment]
sounds good - done

File: src/refiners/foundationals/latent_diffusion/model.py [hunk 2]
```diff
@@ -91,15 +91,17 @@ def forward(
...
-        noise = unconditional_prediction + condition_scale * (conditional_prediction - unconditional_prediction)
+        predicted_noise = unconditional_prediction + condition_scale * (
+            conditional_prediction - unconditional_prediction
+        )
         x = x.narrow(dim=1, start=0, length=4)  # support > 4 channels for inpainting
 
         if self.has_self_attention_guidance():
-            noise += self.compute_self_attention_guidance(
+            predicted_noise += self.compute_self_attention_guidance(
                 x=x, noise=unconditional_prediction, step=step, clip_text_embedding=clip_text_embedding, **kwargs
```
deltheil, line 101: We could have been tempted to do some renaming for `compute_self_attention_guidance` `noise` argument. But let's keep it for later (definitely fine as-is)

deltheil (approved):
Thanks!

src/refiners/foundationals/latent_diffusion/model.py
brycedrennan: Yeah true. 

Also FYI I noticed that self_attention_guidance doesn't work with the euler scheduler. Some typing mismatch that I didn't look into.

deltheil: > FYI I noticed that self_attention_guidance doesn't work with the euler scheduler

Definitely - which is also why it has to be revamped (e.g. `remove_noise` must be implemented differently in this case).

PR #207, opened 2024-01-22
Load Multiple LoRAs with SDLoraManager

limiteinductive:


tests/foundationals/latent_diffusion/test_lora_manager.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_names(manager: SDLoraManager, weights: dict[str, torch.Tensor]) -> None:
+    assert manager.names == []
+
+    manager.add_loras("pokemon-lora", tensors=weights)
+    assert manager.names == ["pokemon-lora"]
+
+    manager.add_loras("pokemon-lora2", tensors=weights)
+    assert manager.names == ["pokemon-lora", "pokemon-lora2"]
```
deltheil, line 78:
```suggestion
    assert set(manager.names) == {"pokemon-lora", "pokemon-lora2"}
```

deltheil: For some reasons, I can't get the output with expected order on my own... (even though it comes from a `list(set(...))` internally which is supposed to maintain insertion order (?)). Please double check on your own.

deltheil (dismissed):
See final comment re: `test_lora_manager.py`

deltheil (approved):

PR #213, opened 2024-01-27
List inputs in lda encode/decode, and clip text_encoder.

piercus:
Naming of `lda.decode_latents(latents)` feels strange cause `lda.decode(latents)` is also decoding latents.

I suggest using `lda.decode_image`/`lda.decode_images` and `lda.encode_image`/`lda.encode_images`

To not break examples, the current PR is keeping an alias `lda.decode_latents(latents) -> lda.decode_image(latents)`

### DIFF
--- a/src/refiners/foundationals/clip/tokenizer.py
+++ b/src/refiners/foundationals/clip/tokenizer.py
@@ -4,7 +4,7 @@ from functools import lru_cache
 from itertools import islice
 from pathlib import Path
 
-from torch import Tensor, tensor
+from torch import Tensor, tensor, cat
 
 import refiners.fluxion.layers as fl
 from refiners.fluxion import pad
@@ -51,11 +51,20 @@ class CLIPTokenizer(fl.Module):
         self.end_of_text_token_id: int = end_of_text_token_id
         self.pad_token_id: int = pad_token_id
 
-    def forward(self, text: str) -> Tensor:
+    def forward(self, text: str | list[str]) -> Tensor:
+        
+        if isinstance(text, str):
+            return self.tokenize_str(text)
+        else:
+            assert isinstance(text, list), f"Expected type `str` or `list[str]`, got {type(text)}"
+            return cat([self.tokenize_str(txt) for txt in text])
+        
+    def tokenize_str(self, text: str) -> Tensor:
         tokens = self.encode(text=text, max_length=self.sequence_length).unsqueeze(dim=0)
+        
         assert (
             tokens.shape[1] <= self.sequence_length
-        ), f"Text is too long: tokens.shape[1] > sequence_length: {tokens.shape[1]} > {self.sequence_length}"
+        ), f"Text is too long: tokens.shape[1] > sequence_length: {tokens.shape[1]} > {self.sequence_length}, {len(text)}"
         return pad(x=tokens, pad=(0, self.sequence_length - tokens.shape[1]), value=self.pad_token_id)
 
     @lru_cache()
--- a/src/refiners/foundationals/latent_diffusion/auto_encoder.py
+++ b/src/refiners/foundationals/latent_diffusion/auto_encoder.py
@@ -1,5 +1,5 @@
 from PIL import Image
-from torch import Tensor, device as Device, dtype as DType
+from torch import Tensor, device as Device, dtype as DType, cat
 
 from refiners.fluxion.context import Contexts
 from refiners.fluxion.layers import (
@@ -211,11 +211,25 @@ class LatentDiffusionAutoencoder(Chain):
         return x
 
     def encode_image(self, image: Image.Image) -> Tensor:
-        x = image_to_tensor(image, device=self.device, dtype=self.dtype)
+        return self.encode_images([image])
+    
+    def encode_images(self, images: list[Image.Image]) -> Tensor:
+        x = cat([image_to_tensor(image, device=self.device, dtype=self.dtype) for image in images], dim=0)
         x = 2 * x - 1
         return self.encode(x)
-
+    
     def decode_latents(self, x: Tensor) -> Image.Image:
+        # For retro-compatibilty
+        return self.decode_image(x)
+    
+    def decode_image(self, x: Tensor) -> Image.Image:
+        if x.shape[0] != 1:
+            raise ValueError(f"Expected batch size of 1, got {x.shape[0]}")
+        
+        return self.decode_images(x)[0]
+        
+    def decode_images(self, x: Tensor) -> list[Image.Image]:
         x = self.decode(x)
         x = (x + 1) / 2
-        return tensor_to_image(x)
+        return [tensor_to_image(t) for t in x.chunk(x.shape[0])]
+        
--- a/tests/foundationals/clip/test_text_encoder.py
+++ b/tests/foundationals/clip/test_text_encoder.py
@@ -101,3 +101,8 @@ def test_encoder(
     # numerical differences depending on the backend.
     # Also we use FP16 weights.
     assert (our_embeddings - ref_embeddings).abs().max() < 0.01
+    
+    # batched inputs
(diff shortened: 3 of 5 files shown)


src/refiners/foundationals/latent_diffusion/auto_encoder.py
```diff
@@ -211,11 +211,24 @@ def decode(self, x: Tensor) -> Tensor:
...
+        if x.shape[0] != 1:
+            raise ValueError(f"Expected batch size of 1, got {x.shape[0]}")
+
+        return self.decode_images(x)[0]
+
+    def decode_images(self, x: Tensor) -> list[Image.Image]:
         x = self.decode(x)
         x = (x + 1) / 2
-        return tensor_to_image(x)
+        return [tensor_to_image(t) for t in x.chunk(x.shape[0])]
```
deltheil, line 234:
```suggestion
        return [tensor_to_image(t) for t in x.split(1)]
```

src/refiners/foundationals/latent_diffusion/auto_encoder.py
deltheil, line 1: Putting backward-compatibility aside, I think another option for these higher-level helpers could be:

```python
def image_to_latent(self, image: Image.Image) -> Tensor:
    ...

def images_to_latents(self, images: list[Image.Image]) -> Tensor:
    ...
```

And symmetrically:

```python
def latent_to_image(self, x: Tensor) -> Image.Image:
    ...

def latents_to_images(self, x: Tensor) -> list[Image.Image]:
    ...
```

That way it is super explicit and there is no confusion with the lower-level counterparts (`encode` and `decode`). WDYT? cc @limiteinductive

src/refiners/foundationals/latent_diffusion/auto_encoder.py
deltheil: nit: feels slightly more expressive

src/refiners/foundationals/latent_diffusion/auto_encoder.py
limiteinductive: stgm, it is always better to be explicit, especially in a flow as intricate as latent diffusion can be sometimes

deltheil: @piercus thanks for these changes! Let us know when the PR is ready for review again.

piercus:
@deltheil @limiteinductive

In current code `latents` and `latent` are both used for a single Tensor.
I want to suggest to switch every single tensor variable to singular `latent` and use `latents` only when there is a sequence. 

The naming for helpers would be `image_to_latent` and `images_to_latent` (no s)

WDYT ?

deltheil:
> In current code `latents` and `latent` are both used for a single Tensor.
> I want to suggest to switch every single tensor variable to singular `latent` and use `latents` only when there is a sequence.

Not sure using plural will mean the same depending on the reader (e.g. latents -> latent variables vs. sequence of latent vectors). Perhaps we could make our life simpler and just adopt "latents" systematically (seems pretty widespread already: "CLIP latents", etc).

@limiteinductive any strong preference on your end?

limiteinductive:
> @limiteinductive, is there any strong preference on your end?

It's almost a convention that "latent" is always plural. Since a Tensor is a batch, you can have a single Tensor is still have multiple "latents."

piercus:
Thanks for your help @deltheil @limiteinductive 
Please review this new version.

src/refiners/fluxion/utils.py
```diff
@@ -113,6 +114,12 @@ def default_sigma(kernel_size: int) -> float:
     return tensor
 
 
+def images_to_tensor(
+    images: List[Image.Image], device: Device | str | None = None, dtype: DType | None = None
+) -> Tensor:
```
deltheil, line 119: Since we target 3.10+, `list[Image.Image]` is enough (and you should drop the `from typing import List` of course):

```python
def images_to_tensor(
    images: list[Image.Image], device: Device | str | None = None, dtype: DType | None = None
) -> Tensor:
```

src/refiners/fluxion/utils.py
```diff
@@ -135,6 +142,10 @@ def image_to_tensor(image: Image.Image, device: Device | str | None = None, dtyp
     return image_tensor.unsqueeze(0)
 
 
+def tensor_to_images(tensor: Tensor) -> List[Image.Image]:
```
deltheil, line 145: ditto: `List` -> `list`

src/refiners/foundationals/clip/tokenizer.py
```diff
@@ -51,11 +51,19 @@ def __init__(
...
+            assert isinstance(text, list), f"Expected type `str` or `list[str]`, got {type(text)}"
+            return cat([self.tokenize_str(txt) for txt in text])
+
+    def tokenize_str(self, text: str) -> Tensor:
         tokens = self.encode(text=text, max_length=self.sequence_length).unsqueeze(dim=0)
+
         assert (
             tokens.shape[1] <= self.sequence_length
-        ), f"Text is too long: tokens.shape[1] > sequence_length: {tokens.shape[1]} > {self.sequence_length}"
+        ), f"Text is too long: tokens.shape[1] > sequence_length: {tokens.shape[1]} > {self.sequence_length}, {len(text)}"
```
deltheil, line 66: Wouldn't it be slightly clearer to have: `f"Text is too long ({len(text)}): ..."`?

tests/e2e/test_diffusion.py
```diff
@@ -648,7 +648,7 @@ def test_diffusion_std_random_init(
             clip_text_embedding=clip_text_embedding,
             condition_scale=7.5,
         )
-    predicted_image = sd15.lda.decode_latents(x)
+    predicted_image = sd15.lda.latent_to_image(x)
```
deltheil, line 651:
```suggestion
    predicted_image = sd15.lda.latents_to_image(x)
```

deltheil: Same applies below: please update the other occurrences

tests/foundationals/clip/test_text_encoder.py
```diff
@@ -101,3 +101,7 @@ def test_encoder(
     # numerical differences depending on the backend.
     # Also we use FP16 weights.
     assert (our_embeddings - ref_embeddings).abs().max() < 0.01
+
+    # batched inputs
+    double_tokens = tokenizer([prompt, prompt[0:3]])
+    assert double_tokens.shape[0] == 2
```
deltheil, line 107: I would turn this into a distinct unit test

tests/foundationals/latent_diffusion/test_auto_encoder.py
```diff
@@ -40,12 +40,18 @@ def sample_image(ref_path: Path) -> Image.Image:
...
     # Ensure no saturation. The green channel (band = 1) must not max out.
     assert max(iter(decoded.getdata(band=1))) < 255  # type: ignore
 
     ensure_similar_images(sample_image, decoded, min_psnr=20, min_ssim=0.9)
+
+    encoded = encoder.images_to_latents([sample_image, sample_image])
+    images = encoder.latents_to_images(encoded)
+    assert isinstance(images, list)
+    assert len(images) == 2
+    ensure_similar_images(sample_image, images[1], min_psnr=20, min_ssim=0.9)
```
deltheil, line 57: ditto: I feel like a distinct unit test would be clearer

deltheil (commented):
Thanks! Only minor feedback, please take a look

deltheil (dismissed):

deltheil:
@piercus do you mind rebasing this branch onto `main` and solve the conflicts? (ideally, you should also squash into a single commit - but I can do it myself at the end if that's simpler).

piercus:
Squash and merge done

deltheil (approved):

PR #259, opened 2024-02-06
Fix github actions triggers

hpouillot:
Always execute lint_and_typecheck jobs but skip steps that are not necessary, ie when sources didn't change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,7 @@ name: CI
 
 on:
   push:
-    paths:
-      - src/**
-      - tests/**
-      - scripts/**
-      - .github/workflows/ci.yml
+    branches: ["**"]
   pull_request_target:
     types: [labeled]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,11 +2,7 @@ name: CI
 
 on:
   push:
-    paths:
-      - src/**
-      - tests/**
-      - scripts/**
-      - .github/workflows/ci.yml
```
deltheil, line 9: What about this one?

hpouillot: It's not needed anymore

deltheil (approved):

.github/workflows/ci.yml
```diff
@@ -2,11 +2,7 @@ name: CI
...
+    branches: ["**"]
```
catwell, line 5: What is the point of adding this? Does it do something different from the default?

hpouillot: yes, the workflow doesn't run on tag push when `branches` is specified

PR #263, opened 2024-02-07
batch inputs for compute_clip_text_embedding

piercus:
## Context

I need this for #165, to do GPU efficient evaluation (batch_size > 1).
This is the extension of #213 
This is also a baby step in the context of #255 @limiteinductive 

### One question

In `compute_clip_image_embedding` there is a `concat_batches` bool, i'm shared if we need to put it in `compute_clip_text_embedding` or not.

2 contradictory POV
* I did not put it now cause the functionnal use case is not clear, 
* For convergence of inputs formats (in the context of #255 SD refactoring), we might want to tend to uniformity here.

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
+++ b/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
@@ -68,7 +68,7 @@ class StableDiffusion_1(LatentDiffusionModel):
             dtype=dtype,
         )
 
-    def compute_clip_text_embedding(self, text: str, negative_text: str = "") -> Tensor:
+    def compute_clip_text_embedding(self, text: str | list[str], negative_text: str | list[str] = "") -> Tensor:
         """Compute the CLIP text embedding associated with the given prompt and negative prompt.
 
         Args:
@@ -76,12 +76,14 @@ class StableDiffusion_1(LatentDiffusionModel):
             negative_text: The negative prompt to compute the CLIP text embedding of.
                 If not provided, the negative prompt is assumed to be empty (i.e., `""`).
         """
-        conditional_embedding = self.clip_text_encoder(text)
-        if text == negative_text:
-            return torch.cat(tensors=(conditional_embedding, conditional_embedding), dim=0)
+        text = [text] if isinstance(text, str) else text
+        negative_text = [negative_text] if isinstance(negative_text, str) else negative_text
+        assert len(text) == len(negative_text), "The length of the text list and negative_text should be the same"
 
-        negative_embedding = self.clip_text_encoder(negative_text or "")
-        return torch.cat(tensors=(negative_embedding, conditional_embedding), dim=0)
+        conditional_embedding = self.clip_text_encoder(text)
+        negative_embedding = self.clip_text_encoder(negative_text)
+
+        return torch.cat((negative_embedding, conditional_embedding))
 
     def set_unet_context(self, *, timestep: Tensor, clip_text_embedding: Tensor, **_: Tensor) -> None:
         """Set the various context parameters required by the U-Net model.
--- a/src/refiners/foundationals/latent_diffusion/stable_diffusion_xl/text_encoder.py
+++ b/src/refiners/foundationals/latent_diffusion/stable_diffusion_xl/text_encoder.py
@@ -1,7 +1,7 @@
 from typing import cast
 
 from jaxtyping import Float
-from torch import Tensor, cat, device as Device, dtype as DType
+from torch import Tensor, cat, device as Device, dtype as DType, split
 
 import refiners.fluxion.layers as fl
 from refiners.fluxion.adapters.adapter import Adapter
@@ -40,7 +40,7 @@ class TextEncoderWithPooling(fl.Chain, Adapter[CLIPTextEncoderG]):
     def init_context(self) -> Contexts:
         return {"text_encoder_pooling": {"end_of_text_index": []}}
 
-    def __call__(self, text: str) -> tuple[Float[Tensor, "1 77 1280"], Float[Tensor, "1 1280"]]:
+    def __call__(self, text: str | list[str]) -> tuple[Float[Tensor, "batch 77 1280"], Float[Tensor, "batch 1280"]]:
         return super().__call__(text)
 
     @property
@@ -48,13 +48,14 @@ class TextEncoderWithPooling(fl.Chain, Adapter[CLIPTextEncoderG]):
         return self.ensure_find(CLIPTokenizer)
 
     def set_end_of_text_index(self, end_of_text_index: list[int], tokens: Tensor) -> None:
-        position = (tokens == self.tokenizer.end_of_text_token_id).nonzero(as_tuple=True)[1].item()
-        end_of_text_index.append(cast(int, position))
+        for str_tokens in split(tokens, 1):
+            position = (str_tokens == self.tokenizer.end_of_text_token_id).nonzero(as_tuple=True)[1].item()  # type: ignore
+            end_of_text_index.append(cast(int, position))
 
-    def pool(self, x: Float[Tensor, "1 77 1280"]) -> Float[Tensor, "1 1280"]:
+    def pool(self, x: Float[Tensor, "batch 77 1280"]) -> Float[Tensor, "batch 1280"]:
         end_of_text_index = self.use_context(context_name="text_encoder_pooling").get("end_of_text_index", [])
-        assert len(end_of_text_index) == 1, "End of text index not found."
-        return x[:, end_of_text_index[0], :]
+        assert len(end_of_text_index) == x.shape[0], "End of text index not found."
+        return cat([x[i : i + 1, end_of_text_index[i], :] for i in range(x.shape[0])], dim=0)
 
 
 class DoubleTextEncoder(fl.Chain):
@@ -75,7 +76,7 @@ class DoubleTextEncoder(fl.Chain):
         tep = TextEncoderWithPooling(target=text_encoder_g, projection=projection)
         tep.inject(self.layer("Parallel", fl.Parallel))
 
-    def __call__(self, text: str) -> tuple[Float[Tensor, "1 77 2048"], Float[Tensor, "1 1280"]]:
+    def __call__(self, text: str | list[str]) -> tuple[Float[Tensor, "batch 77 2048"], Float[Tensor, "batch 1280"]]:
         return super().__call__(text)
 
     def concatenate_embeddings(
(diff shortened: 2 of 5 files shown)


piercus:
@limiteinductive just added corresponding changes + unit tests for SDXL

src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
```diff
@@ -78,10 +78,19 @@ def compute_clip_text_embedding(self, text: str, negative_text: str = "") -> Ten
...
+                ), "The length of the text list and negative_text should be the same"
+
+            if isinstance(negative_text, str) and isinstance(text, list):
+                negative_text = [negative_text] * len(text)
+
+            negative_embedding = self.clip_text_encoder(negative_text)
 
-        negative_embedding = self.clip_text_encoder(negative_text or "")
-        return torch.cat(tensors=(negative_embedding, conditional_embedding), dim=0)
+        return cat((negative_embedding, conditional_embedding))
```
limiteinductive, line 93: Let's make it more compact:
```py
        text = [text] if isinstance(text, str) else text
        negative_text = [negative_text] if isinstance(negative_text, str) else negative_text
        assert len(text) == len(negative_text), "The length of the text list and negative_text should be the same"
        conditional_embedding = self.clip_text_encoder(text)
        negative_embedding = self.clip_text_encoder(negative_text) if negative_text else conditional_embedding
        return cat((negative_embedding, conditional_embedding))
 ```
(or smthg like that)

**Comment by deltheil** [comment]
> In `compute_clip_image_embedding` there is a `concat_batches` bool, i'm shared if we need to put it in `compute_clip_text_embedding` or not.

@piercus no, you can ignore it IMO. It has been added mainly for documentation purposes, in the context of IP-Adapter with multiple image prompts (see #218). And in this context, you want to create a longer sequence of tokens (like double the size for two images).

File: src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py (hunk 1 above)
Comment (piercus) [reply src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py]: done

**Review by limiteinductive — DISMISSED** [review]
@deltheil lgtm

**Comment by deltheil** [comment]
> @deltheil lgtm

Thanks! @piercus will do the final round shortly, stay tuned

File: src/refiners/foundationals/latent_diffusion/stable_diffusion_xl/text_encoder.py [hunk 2]
```diff
@@ -40,21 +40,22 @@ def __init__(
...
         return super().__call__(text)
 
     @property
     def tokenizer(self) -> CLIPTokenizer:
         return self.ensure_find(CLIPTokenizer)
 
     def set_end_of_text_index(self, end_of_text_index: list[int], tokens: Tensor) -> None:
-        position = (tokens == self.tokenizer.end_of_text_token_id).nonzero(as_tuple=True)[1].item()
-        end_of_text_index.append(cast(int, position))
+        for str_tokens in tokens.split(1):  # type: ignore
```
deltheil, line 51: Let's add `split` to `torch` imports and replace with:

```python
for str_tokens in split(tokens, 1):
    ...
```
That way you can get rid of the `# type: ignore`

tests/foundationals/latent_diffusion/test_sdxl_double_encoder.py
```diff
@@ -100,3 +100,24 @@ def test_double_text_encoder(diffusers_sdxl: DiffusersSDXL, double_text_encoder:
 
     assert torch.allclose(input=negative_double_embedding, other=negative_prompt_embeds, rtol=1e-3, atol=1e-3)
     assert torch.allclose(input=negative_pooled_embedding, other=negative_pooled_prompt_embeds, rtol=1e-3, atol=1e-3)
+
+
+@no_grad()
+def test_double_text_encoder_batch2(diffusers_sdxl: DiffusersSDXL, double_text_encoder: DoubleTextEncoder) -> None:
```
deltheil, line 106:
```suggestion
def test_double_text_encoder_batch2(double_text_encoder: DoubleTextEncoder) -> None:
```

deltheil: Unused

tests/e2e/test_diffusion.py
```diff
@@ -1106,6 +1138,68 @@ def test_diffusion_lora(
     ensure_similar_images(predicted_image, expected_image, min_psnr=35, min_ssim=0.98)
 
 
+@no_grad()
+def test_diffusion_sdxl_batch2(sdxl_ddim: StableDiffusion_XL) -> None:
```
deltheil, line 1142: Any reason why you did not write `test_diffusion_batch2` (for SD1) the same way? i.e. just one diffusion step + comparison with/without batching?

tests/e2e/test_diffusion.py
```diff
@@ -1106,6 +1138,68 @@ def test_diffusion_lora(
...
+        step=sdxl.steps[0],
+        clip_text_embedding=clip_text_embedding_2,
+        pooled_text_embedding=pooled_text_embedding_2,
+        time_ids=time_ids,
+    )
+
+    predicted_image_2 = sdxl.lda.latents_to_image(x_2)
+
+    ensure_similar_images(predicted_image_b2[0], predicted_image_1, min_psnr=35, min_ssim=0.98)
+    ensure_similar_images(predicted_image_b2[1], predicted_image_2, min_psnr=35, min_ssim=0.98)
```
deltheil, line 1200: I was thinking that, perhaps, given you only perform one diffusion step, it'd be simpler just to compare the noise predictions (torch allclose) and not VAE decode. WDYT?

deltheil (commented):
Some comments/suggestions, please take a look

tests/e2e/test_diffusion.py
piercus: It would makes sense to compare latents directly.

Not sure about comparing noise.
What we want to test here is e2e "batch-stability" and then including the noise -> latents solver/cfg logic in the test makes sense.

WDYT ?

piercus: FYI, switching to `torch.allclose` makes the test fail

```python
torch.norm(x_b2[1] - x_2[0])
=> 
tensor(1.4638, device='cuda:0')
```

Studying this, it might take some time

piercus:
The batch-stability on GPU suffers from the below effect

## Source : Torch Double Conv2d

In the above example, a pure pytorch double Conv2d leads to 2e-4 differance

```
import torch
from torch.nn import Conv2d 
device = "cuda:0"

with torch.no_grad():
    torch.cuda.manual_seed_all(0)
    x_b2 = torch.randn(2, 4, 32, 32).to(device)
    conv2d_1 = Conv2d(in_channels=4, out_channels=320, kernel_size=3, padding=1, device=device)
    conv2d_2 = Conv2d(in_channels=320, out_channels=640, kernel_size=3, padding=1, device=device)
    print((output_b2[0] - output_b1[0]).norm())
```
## Amplification

This batch discrepancy is then amplified by the following layers  

| Level | Error `(output_b2[0] - output_b1[0]).norm()` |
|--|--|
| Double Conv | 2e-4 |
| Conv + ResidualBlock | 7e-3 |
| DownBlocks | 7e-2 |
| Unet | 1e-2 | 
| SD | 3.3e-2 | 

Details of the analysis can be found in https://gist.github.com/piercus/07d03f258907542d312c0c735445e793

## Result

As a results the batch stability with `torch.allclose` is only calculated at 5e-2, which can be considered as quite weak test

tests/e2e/test_diffusion.py
piercus: 1.46 was a bug, but still need to put absolute tolerance to 5e-3 (which seems quite high) see https://github.com/finegrain-ai/refiners/pull/263#issuecomment-1956404911 for the details

tests/e2e/test_diffusion.py
piercus: no good reasons, fixed now

tests/e2e/test_diffusion.py
```diff
@@ -757,6 +757,63 @@ def test_diffusion_std_random_init(
...
+
+    clip_text_embedding_b2 = sd15.compute_clip_text_embedding(
+        text=[prompt1, prompt2], negative_text=[negative_prompt1, negative_prompt2]
+    )
+
+    sd15.set_inference_steps(30)
+    step = sd15.steps[0]
+
+    manual_seed(2)
+    rand_b2 = torch.randn(2, 4, 64, 64, device=test_device)
```
deltheil, line 779:
```suggestion
def test_diffusion_batch2(sd15_std: StableDiffusion_1):
    sd15 = sd15_std

    prompt1 = "a cute cat, detailed high-quality professional image"
    negative_prompt1 = "lowres, bad anatomy, bad hands, cropped, worst quality"
    prompt2 = "a cute dog"
    negative_prompt2 = "lowres, bad anatomy, bad hands"

    clip_text_embedding_b2 = sd15.compute_clip_text_embedding(
        text=[prompt1, prompt2], negative_text=[negative_prompt1, negative_prompt2]
    )

    step = sd15.steps[0]

    manual_seed(2)
    rand_b2 = torch.randn(2, 4, 64, 64, device=sd15.device)
```

deltheil: To mimic `test_diffusion_sdxl_batch2` (and `expected_image_std_random_init` is unused anyway). Also, got rid of `set_inference_steps` which is not needed here.

tests/e2e/test_diffusion.py
```diff
@@ -1265,6 +1321,69 @@ def test_diffusion_lora(
...
+    prompt2 = "a cute dog"
+    negative_prompt2 = "lowres, bad anatomy, bad hands"
+
+    clip_text_embedding_b2, pooled_text_embedding_b2 = sdxl.compute_clip_text_embedding(
+        text=[prompt1, prompt2], negative_text=[negative_prompt1, negative_prompt2]
+    )
+
+    time_ids = sdxl.default_time_ids
+    time_ids_b2 = sdxl.default_time_ids.repeat(2, 1)
+    sdxl.set_inference_steps(40)
```
deltheil, line 1339:
```suggestion
```

deltheil: Given what this test does, there is no need to change the number of inference steps, right?

deltheil (approved):
See final nits. Please squash the extra commit afterwards. Thanks!

piercus:
@deltheil please review

deltheil (approved):

PR #264, opened 2024-02-07
Less than 1 epoch training duration

piercus:


limiteinductive:
Could you rebase and then I think we can merge this?

limiteinductive (commented):
Needs to be rebased

piercus:
@limiteinductive rebase done

limiteinductive (dismissed):

src/refiners/training_utils/trainer.py
```diff
@@ -522,9 +522,12 @@ def step(self, batch: Batch) -> None:
...
         for batch in self.dataloader:
-            self._call_callbacks(event_name="on_batch_begin")
-            self.step(batch=batch)
-            self._call_callbacks(event_name="on_batch_end")
+            if not self.clock.done:
+                self._call_callbacks(event_name="on_batch_begin")
+                self.step(batch=batch)
+                self._call_callbacks(event_name="on_batch_end")
+            else:
+                break
```
deltheil, line 530:
```suggestion
            if self.clock.done:
                break
            self._call_callbacks(event_name="on_batch_begin")
            self.step(batch=batch)
            self._call_callbacks(event_name="on_batch_end")
```

deltheil: nit: would look slightly clearer like this IMO

deltheil (changes requested):
@piercus just a small nit, PTAL

deltheil (approved):

PR #266, opened 2024-02-07
do not regenerate lock files

hpouillot:
speedup installation by not generating lock files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: rye sync
         if: steps.changes.outputs.sources == 'true'
-        run: rye sync --all-features
+        run: rye sync --no-lock --all-features
 
       - name: ruff format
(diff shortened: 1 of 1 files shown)


deltheil (approved):

catwell:
I think this won't work w/o `--no-dev` since we don't commit the dev lockfile, you either need both or none.

catwell (changes requested):
I think this won't work w/o --no-dev since we don't commit the dev lockfile, you either need both or none.

hpouillot:
Indeed, didn't know the dev lockfile wasn't part of the repo.

catwell:
@limiteinductive @deltheil WDYT?

In Disciple or PHX I don't do it because things like pyright are not dev deps so it is useless (and I can use dev to prototype stuff locally). But here if we want to use the dev deps and not generate the lock we'll have to do this.

On the other hand if we do this we need to remember to update the lock manually, we won't detect breaking changes in deps as frequently.

PR #268, opened 2024-02-07
Make Dataset part of Trainer

limiteinductive:


src/refiners/training_utils/trainer.py
```diff
@@ -250,6 +250,23 @@ def compute_grad_norm(parameters: Iterable[Parameter]) -> float:
 ConfigType = TypeVar("ConfigType", bound=BaseConfig)
 
 
+class _Dataset(Dataset[Batch]):
+    """
+    A wrapper around the `get_item` method to create a PyTorch Dataset.
```
deltheil, line 255:
```suggestion
    A wrapper around the `get_item` method to create a [`torch.utils.data.Dataset`][torch.utils.data.Dataset].
```

deltheil (approved):
Please see minor suggestion re: docstring

PR #272, opened 2024-02-08
Add `typos` and `rich` to dev-dependencies

Laurent2916:
I added `typos` to the dev-deps, as it is now in the CI. I also added `rich` to the dev-deps (as I personally use it quite a lot when developing).

I also noticed that `ruff` was in the normal dependencies (and not only in the dev-dependencies, so I removed it from the dependencies section and from requirements.lock).

(other diffs are from my auto-formatter)


pyproject.toml
```diff
@@ -70,10 +69,12 @@ build-backend = "hatchling.build"
 managed = true
 dev-dependencies = [
     "pyright==1.1.349",
-    "ruff>=0.1.15",
+    "ruff>=0.2.0",
     "docformatter>=1.7.5",
     "pytest>=8.0.0",
     "coverage>=7.4.1",
+    "typos>=1.18.2",
+    "rich>=13.7.0",
```
deltheil, line 77: Perhaps you should just install it as global tool -> https://rye-up.com/guide/tools/

catwell: I agree, adding ruff as a dependency instead of a dev dependency was a mistake and adding `typos` makes sense since it's part of the CI, but things like `rich` are personal preferences and should not go in the dev deps of the project.

Laurent2916: Didn't know about global tools, I'll do that.

deltheil (approved):

PR #274, opened 2024-02-09
feature: Implement Consistency decoder for SDv1VAE

rodSiry:
- [x] Check if the original implementation of consistency decoder decodes an image encoded from refiners SD1Autoencoder() 
- [x] Identify model architecture
- [x] Write chain module classes
- [x] Functional weight converter
- [x] Outer decoding loop & final test
- [ ] cleaning, refactoring, unit tests...

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/solvers/consistency_decoder_solver.py
+++ b/src/refiners/foundationals/latent_diffusion/solvers/consistency_decoder_solver.py
@@ -1,0 +1,140 @@
+import math
+import torch
+from torch import Generator, Tensor, device as Device, dtype as Dtype, float32
+from typing import List
+from refiners.foundationals.latent_diffusion.solvers.solver import NoiseSchedule, Solver
+from refiners.foundationals.latent_diffusion.solvers.solver import SolverParams
+
+
+# Copied from diffusers.schedulers.scheduling_ddpm.betas_for_alpha_bar
+def betas_for_alpha_bar(
+    num_diffusion_timesteps: int,
+    max_beta :float =0.999,
+    alpha_transform_type: str="cosine",
+):
+    """
+    Create a beta schedule that discretizes the given alpha_t_bar function, which defines the cumulative product of
+    (1-beta) over time from t = [0,1].
+
+    Contains a function alpha_bar that takes an argument t and transforms it to the cumulative product of (1-beta) up
+    to that part of the diffusion process.
+
+
+    Args:
+        num_diffusion_timesteps (`int`): the number of betas to produce.
+        max_beta (`float`): the maximum beta to use; use values lower than 1 to
+                     prevent singularities.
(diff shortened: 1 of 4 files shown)


rodSiry:
I think the feature is functional (results below)

Before further cleaning, I'd like to know : 
-Should I define a ConsistencyAutoencoder class (i.e. default encoder + consistency decoder) ?

Don't hesitate to provide additional feedback.

ground truth
Ground Truth
default decoder
Default decoder
consistency decoder (diffusers)
Consistency decoder (from diffusers)
consistency decoder (ours)
Consistency decoder (from refiners)

src/refiners/foundationals/latent_diffusion/solvers/consistency_decoder_solver.py
```diff
@@ -0,0 +1,149 @@
...
+        raise ValueError(f"Unsupported alpha_tranform_type: {alpha_transform_type}")
+
+    betas = []
+    for i in range(num_diffusion_timesteps):
+        t1 = i / num_diffusion_timesteps
+        t2 = (i + 1) / num_diffusion_timesteps
+        betas.append(min(1 - alpha_bar_fn(t2) / alpha_bar_fn(t1), max_beta))
+    return torch.tensor(betas, dtype=torch.float32)
+
+class ConsistencyDecoderSolver(Solver):
```
limiteinductive, line 52: Is it possible to use instead the existing EulerSolver? If not, what are the differences?

limiteinductive (commented):
Thanks for the submission! I have been able to make the conversion works on my side. Could you rebase your branch / remove unecessary files / fix pyright issues. Then two main question for the moment:

 1/ Can we use the existing Euler solver?
 2/ Could you write an e2e test? This will also help me review by having a snippet for inference.

src/refiners/foundationals/latent_diffusion/solvers/consistency_decoder_solver.py
rodSiry: The ConsistencyDecoderVAE comes with a custom, specific scheduler class in diffusers.
I'll double check to see if it boils down to Euler with well chosen inference parameters.

limiteinductive: Thanks. In any case, you should implement ConsistencyDecoder using Refiners terminology for solvers (we don't use greek letters)

catwell: :wave: note that we changed the interface of solvers a bit in 0.4 (https://github.com/finegrain-ai/refiners/pull/305) so you will need to rebase and change a few things.

rodSiry: Hi, 

Sorry for the wait, had a week off, I did the rebase. 

Will re-work the solver and do the e2e tests asap

PR #278, opened 2024-02-10
feature: Record batch time

isamu-isozaki:
This piece of code helped me identify how much time is spent in the data processing stage, the forward pass, and backward pass so that I know where to optimize.
The main steps blocking this pr will be

- [x] Figure out how to log with trainer since I think the def log function is not present currently because of refactoring


isamu-isozaki:
Now that I think about it, I think this can be a callback

isamu-isozaki:
I'm thinking if this can become a callback

src/refiners/training_utils/trainer.py
```diff
@@ -356,25 +379,39 @@ def backward(self) -> None:
...
         loss = self.compute_loss(batch=batch)
         self.loss = loss
+        forward_time = time.time() - start
+        self.forward_time_m.update(forward_time)
+        start = time.time()
         self._call_callbacks(event_name="on_compute_loss_end")
-        self.backward()
+        backward_time = self.backward(start)
+        self.backprop_time_m.update(backward_time)
+        return forward_time, backward_time
```
limiteinductive, line 399: Instead of making the signature of `step` more complex, maybe we could use a decorator and attributes to leave it as is

src/refiners/training_utils/trainer.py
```diff
@@ -141,6 +160,10 @@ def __init__(self, config: ConfigType) -> None:
         self._models: ModelRegistry = {}
         self._callbacks: CallbackRegistry = {}
         self.config = config
+        self.batch_time_m = AverageMeter()
```
limiteinductive, line 163: nit: What does `m` stand for? I prefer to use explicit names (even if verbose).

src/refiners/fluxion/utils.py
```diff
@@ -229,8 +229,7 @@ def load_tensors(path: Path | str, /, device: Device | str = "cpu") -> dict[str,
         tensors = torch.load(path, map_location=device, weights_only=True)  # type: ignore
 
     assert isinstance(tensors, dict) and all(
-        isinstance(key, str) and isinstance(value, Tensor)
-        for key, value in tensors.items()  # type: ignore
+        isinstance(key, str) and isinstance(value, Tensor) for key, value in tensors.items()  # type: ignore
```
limiteinductive, line 232: not related to PR

src/refiners/training_utils/trainer.py
isamu-isozaki: ah meter. Yeah let me fix that

src/refiners/foundationals/latent_diffusion/stable_diffusion_1/unet.py
```diff
@@ -165,7 +165,8 @@ def __init__(self, device: Device | str | None = None, dtype: DType | None = Non
 class SD1UNet(fl.Chain):
     """Stable Diffusion 1.5 U-Net.
 
-    See [[arXiv:2112.10752] High-Resolution Image Synthesis with Latent Diffusion Models](https://arxiv.org/abs/2112.10752) for more details."""
+    See [[arXiv:2112.10752] High-Resolution Image Synthesis with Latent Diffusion Models](https://arxiv.org/abs/2112.10752) for more details.
```
limiteinductive, line 168: ditto

limiteinductive (changes requested):
Thanks this is gonna be useful! Maybe you can take inspiration on how the TrainingClock Callback is implemented? Let's try to pollute as little as possible the core of the Trainer

isamu-isozaki:
@limiteinductive do you think I should make a timer config or merge timer with clock?

limiteinductive:
> @limiteinductive do you think I should make a timer config or merge timer with clock?

I think it makes sense to merge with Clock and extend its config.

Laurent2916:
Closing this PR as it’s been sitting here for a long time and is unlikely to be merged. Thanks for the effort!

PR #290, opened 2024-02-16
bug/feature: Fixing mixed precision training

isamu-isozaki:
This makes it so we can do lower precision training. @piercus might be interested


src/refiners/training_utils/trainer.py
```diff
@@ -345,12 +349,25 @@ def backward(self) -> None:
         """Backward pass on the loss."""
         self._call_callbacks(event_name="on_backward_begin")
         scaled_loss = self.loss / self.clock.num_step_per_iteration
+        if self.scaler is not None:
+            self.scaler.scale(scaled_loss).backward()  # type: ignore
+        else:
+            backward(tensors=scaled_loss)
         backward(tensors=scaled_loss)
```
limiteinductive, line 356: Don't you have a double backward here?

isamu-isozaki: @limiteinductive oh yup lol let me fix that in a sec

isamu-isozaki: Fixed i think

src/refiners/training_utils/trainer.py
```diff
@@ -345,11 +349,24 @@ def backward(self) -> None:
...
-        backward(tensors=scaled_loss)
+        if self.scaler is not None:
+            self.scaler.scale(scaled_loss).backward()  # type: ignore
+        else:
+            backward(tensors=scaled_loss)
         self._call_callbacks(event_name="on_backward_end")
         if self.clock.is_optimizer_step:
             self._call_callbacks(event_name="on_optimizer_step_begin")
-            self.optimizer.step()
+            if self.scaler is not None:
```
limiteinductive, line 359: My rule is that you should try not to fold code more than 3 times.

In this case, you can write a separate method `optimizer_step`

```py
def optimizer_step(self) -> None:
    if self.scaler is None:
        self.optimizer.step()
        return
    # rest of the logic here
```

limiteinductive (approved):
@deltheil PTAL. @isamu-isozaki could you rebase?

limiteinductive:
When trying with bfloat16 I've got this error
image

limiteinductive:
image
even with float16 I have an error

limiteinductive (changes requested):
See runtime bugs found

isamu-isozaki:
@limiteinductive Ok, a slight update. I was wrong. For bfloat16 it doesn't seem like grad scaling is necessary because the number of exponent bits is the exact same as float32. Though in fp16 it is necessary. The steps needed to make this work for float16 seem like, traditionally,
1. Keep the model in fp32 precision
2. Have a scaler
3. Autocast to fp16 during training
For autocasting usually it's done like
```
with autocast(device_type='cuda', dtype=torch.float16):
```
but in accelerate they do
```
if self.mixed_precision == "fp16" and is_torch_version(">=", "1.10"):
    model.forward = torch.cuda.amp.autocast(dtype=torch.float16)(model.forward)
```
and then do 
```
model.forward = convert_outputs_to_fp32(model.forward)
```
which I think I don't get fully why they are equivalent. Anyway I'll update pr so at least in bf16 it works

isamu-isozaki:
Now that I think about it, even in bfloat16 the gradients should be fp32

isamu-isozaki:
Ok fixed. For this training, the models getting trained have to be in fp32 while all other models can be self.dtype.
The reason I modified the DTypes in the pipelines is that they automatically convert the training model's fp32 to the self.dtype.
For the training step and evaluation, I think we have to do autocast and my guess is it's the simplest way.
In my dino pr, these changes made the gradients fp32 while training on bfloat16

src/refiners/foundationals/latent_diffusion/model.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         clip_text_encoder: fl.Chain,
         solver: Solver,
         device: Device | str = "cpu",
-        dtype: DType = torch.float32,
+        dtype: DType | None = torch.float32,
```
limiteinductive, line 23: We shouldn't change `foundationals` API in a PR about training_utils (they must be flexible enough to accommodate).

src/refiners/training_utils/config.py
```diff
@@ -155,6 +155,7 @@ class ModelConfig(BaseModel):
     betas: tuple[float, float] | None = None
     eps: float | None = None
     weight_decay: float | None = None
+    train: bool = False
```
limiteinductive, line 158: I don't like this "train" attribute ; even when set to False, the model will still train since the learnable_parameter registration works anyway

limiteinductive (changes requested):
It starts to be a bit chaotic. We're adding a lot of API changes to accommodate for one specific scenario, i.e., mix precision training. Here is some proposition to make it better:

 -> Add an optional dtype device attribute in ModelConfig to deal with cases where
 -> Deal with mixed precision training as if it were a dtype (called "amp" ?)

So for instance,

```py
TrainingConfig(
  dtype="amp"
)

MyModelConfig(
   # dtype="amp" I commented it since it's a given that the model will have the dtype of the trainer
)

OtheModelConfig(
 dtype="bfloat16" # and for a model we don't have to compute grad on, we can give the precision we want
)
)
```

src/refiners/foundationals/latent_diffusion/model.py
isamu-isozaki: Hmm, is there a way to initialize a StableDiffusion_1 or XL without setting a Dtype? It can prob be a subject of a different pr if needed

isamu-isozaki:
@limiteinductive Sorry I didn't get the config change but I think if I just set the Dtype of the model if it doesn't have learnable params, then I won't have to modify configs!

limiteinductive:
> @limiteinductive Sorry I didn't get the config change but I think if I just set the Dtype of the model if it doesn't have learnable params, then I won't have to modify configs! alternatively, I can set the learnable parameters to always be float32

Setting just the learnable parameters to float32 if the amp is activated is a good idea.

isamu-isozaki:
Do you think it matters if the amp is active or not? 
I was thinking we will always want learnable parameters=float32

limiteinductive:
> Do you think it matters if the amp is active or not? I was thinking we will always want learnable parameters=float32

This defeats the purpose of low-precision training in most cases, isn't it?

limiteinductive:
> Do you think it matters if the amp is active or not? I was thinking we will always want learnable parameters=float32

So the question is, do we want to force ppl to do mix precision? Or is it always suboptimal to fully train in bfloat16 for instance? I know that autocast can have some serious performance impact.

limiteinductive:
> > Do you think it matters if the amp is active or not? I was thinking we will always want learnable parameters=float32
> 
> So the question is, do we want to force ppl to do mix precision? Or is it always suboptimal to fully train in bfloat16 for instance? I know that autocast can have some serious performance impact.

I think it would be complicated to say that "mixed-precision is universally better than simple low-precision training," and therefore, we should leave the choice of either having a "bfloat16" training and activating or not mixed-precision.

isamu-isozaki:
> > Do you think it matters if the amp is active or not? I was thinking we will always want learnable parameters=float32
> 
> So the question is, do we want to force ppl to do mix precision? Or is it always suboptimal to fully train in bfloat16 for instance? I know that autocast can have some serious performance impact.

Great question. I assumed so and yes auto-casting and just having float32 gradients in general speeds down performance in this run by around 14%.
And got it. I think it'll be hard to disprove but in accelerate it's done that way I believe.
I'm not sure how to make the current dtype work with an option of amp so I'll make a new boolean in training where if amp is set to True, it's the current logic + autocast. And otherwise it sets everything to current dtype(except in fp16 where it'll raise an exception)

src/refiners/training_utils/trainer.py
```diff
@@ -101,10 +102,19 @@ def decorator(func: Callable[[Any, ModelConfigT], ModuleT]) -> ModuleT:
...
             model = func(self, config)
-            model = model.to(self.device, dtype=self.dtype)
+            amp: bool = self.config.training.amp
+            model = model.to(self.device)
             if config.requires_grad is not None:
                 model.requires_grad_(requires_grad=config.requires_grad)
             learnable_parameters = [param for param in model.parameters() if param.requires_grad]
+            model_trained = len(learnable_parameters) > 0
+            if not amp or not model_trained:
+                # training is not amp or model is not trained
```
limiteinductive, line 112: nit: I don't think this comment is needed

src/refiners/training_utils/trainer.py
```diff
@@ -101,10 +102,19 @@ def decorator(func: Callable[[Any, ModelConfigT], ModuleT]) -> ModuleT:
...
+                model = model.to(dtype=self.dtype)
```
limiteinductive, line 113: in any case you wan't to cast to model to dtype, no need for if/else

src/refiners/training_utils/trainer.py
```diff
@@ -341,15 +361,34 @@ def compute_loss(self, batch: Batch) -> Tensor:
...
+    def optimizer_step(self) -> None:
+        if self.scaler is None:
+            self.optimizer.step()
+            return
+        scale_before = self.scaler.get_scale()  # type: ignore
+        self.scaler.step(self.optimizer)  # type: ignore
+        self.scaler.update()  # type: ignore
+        scale_after = self.scaler.get_scale()  # type: ignore
+        # If we reduced the loss scale, it means the optimizer step was skipped because of gradient overflow.
+        if scale_after < scale_before:
```
limiteinductive, line 380: Not sure what this custom logic from accelerate does exactly, let's remove it for now (I mean scale_before, scale_after)

limiteinductive (changes requested):

limiteinductive (dismissed):
@deltheil PTAL

src/refiners/training_utils/trainer.py
```diff
@@ -105,6 +106,9 @@ def wrapper(self: Trainer[BaseConfig, Any], config: ModelConfigT) -> fl.Module:
             if config.requires_grad is not None:
                 model.requires_grad_(requires_grad=config.requires_grad)
             learnable_parameters = [param for param in model.parameters() if param.requires_grad]
+            if self.config.training.automatic_mixed_precision:
+                for learnable_parameter in learnable_parameters:
+                    learnable_parameter.to(dtype=float32)
```
deltheil, line 111: I guess this is because with AMP some layers/ops behave better (range) with float32, right? Doesn't this deserve a comment?

src/refiners/training_utils/config.py
```diff
@@ -22,6 +22,9 @@
 
 class TrainingConfig(BaseModel):
     device: str = "cpu"
+    automatic_mixed_precision: bool = (
+        True  # Enables automatic mixed precision which allows float32 gradients while working with lower precision.
+    )
     dtype: str = "float32"
```
deltheil, line 28: Since AMP is on by default (opt-out), and given that this `dtype` config is used jointly given the `autocast(dtype=self.dtype, enabled=self.config.training.automatic_mixed_precision)` call, any reason not to pick a sane default here? Namely: either `float16` or `bfloat16`. Might also deserve a comment.

(accelerate even offers to "Choose from ‘no’,‘fp16’,‘bf16 or ‘fp8’.")

src/refiners/training_utils/trainer.py
```diff
@@ -341,15 +351,28 @@ def compute_loss(self, batch: Batch) -> Tensor:
     def compute_evaluation(self) -> None:
         pass
 
+    def backward_step(self, scaled_loss: Tensor) -> None:
+        if self.scaler is None:
+            backward(tensors=scaled_loss)
+            return
+        self.scaler.scale(scaled_loss).backward()  # type: ignore
```
deltheil, line 358: Note: as discussed offline, a follow up PR will be needed to properly unscale the gradients before gradient clipping (which happens in between the backward step and optimizer step, as per https://pytorch.org/docs/stable/notes/amp_examples.html#working-with-unscaled-gradients)

But also, what about gradient accumulation? https://pytorch.org/docs/stable/notes/amp_examples.html#working-with-unscaled-gradients

deltheil (commented):
Some comments, PTAL cc @limiteinductive

src/refiners/training_utils/config.py
limiteinductive: We could set it to False by default, and even though float32 + amp doesn't do much, I don't see this as a big issue; I would rather keep the API straightforward.

isamu-isozaki:
I found that if we do
```python
# For all parameters we train in automatic mixed precision we want them to be in float32.
for learnable_parameter in learnable_parameters:
    learnable_parameter.to(dtype=float32)
```
while it's more clear it doesn't actually set the dtype of all the parameters to float32. So, the correct way to do it is to set the parameter dtype in the model before register_model(when requires_grad is set), and then if not amp, set grad to self.dtype I think.

Laurent2916:
Closing this PR as it’s been sitting here for a long time and is unlikely to be merged. Thanks for the effort!

PR #291, opened 2024-02-16
bug: Fixing grad accumulation

isamu-isozaki:
In the current logic, I believe we are just scaling the loss and are not accumulating the gradients. This should fix this. But I will double-check

### DIFF
--- a/src/refiners/training_utils/trainer.py
+++ b/src/refiners/training_utils/trainer.py
@@ -145,7 +145,6 @@ class Trainer(Generic[ConfigType, Batch], ABC):
         self._call_callbacks(event_name="on_init_begin")
         self._load_models()
         self._call_callbacks(event_name="on_init_end")
-
     @register_callback()
     def clock(self, config: ClockConfig) -> TrainingClock:
         return TrainingClock(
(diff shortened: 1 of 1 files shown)


src/refiners/training_utils/trainer.py
```diff
@@ -346,18 +346,19 @@ def backward(self) -> None:
...
-            self.optimizer.step()
-            self.optimizer.zero_grad()
-            self._call_callbacks(event_name="on_optimizer_step_end")
-        if self.clock.is_lr_scheduler_step:
-            self._call_callbacks(event_name="on_lr_scheduler_step_begin")
-            self.lr_scheduler.step()
-            self._call_callbacks(event_name="on_lr_scheduler_step_end")
-        if self.clock.is_evaluation_step:
-            self.evaluate()
+        if self.global_step % self.clock.num_step_per_iteration == 0:
```
limiteinductive, line 349: Isn't that what self.clock.is_optimizer_step is checking?

src/refiners/training_utils/trainer.py
```diff
@@ -369,12 +370,14 @@ def step(self, batch: Batch) -> None:
 
     def epoch(self) -> None:
         """Perform a single epoch."""
+        self.global_step = 1
```
limiteinductive, line 373: Isn't this the same as sefl.clock.epoch?

src/refiners/training_utils/trainer.py
isamu-isozaki: ah, you are right. I think this pr is already implemented then. Closing now

PR #302, opened 2024-02-22
[Ready For Review] Implement Fuyu

LouisRouss:
This is an early draft PR to indicate that i'm working on the fuyu 8b implementation. 
A lot of work still needs to be done, including

-  [x] Finalization of the Fuyu architecture
-  [x] Adaptation of the weights + validate the architecture
-  [x] Change Tokenizer
-  [x] Inference function
-  [x] Unit Testing

I will update this PR and ping you as soon as more significant progress has been made or if I encounter any blockers that require discussion.
Thank you !


LouisRouss:
Hi 
Just to give a quick update, weights now adapt nicely and the model works as intended.  It took me a while to debug things due to my low amount of computational power but everything should be quicker now ! 
Next steps are cleaning and documenting the code, writing a proper inference function and finally some unit testing as written in the original message
Thanks !

src/refiners/foundationals/fuyu/tokenizer.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class FuyuTokenizer(fl.Module):
+    """
+    Implement a Unigram Tokenizer based on a vocabulary file for Fuyu
+    """
+
+    def __init__(
+        self,
+        vocabulary_path: str | Path = Path(__file__).resolve().parent / "tokenizer.json.gz",
```
catwell, line 20: Not reviewing everything, but this is a no-go, we will not add a 3 MB `json.gz` file to the source tree. This should be an asset that we deal with like the weights (downloaded externally).

LouisRouss: Understood, no problem i will change it this evening. Wanted to use the same logic as the CLIP tokenizer and the bpe_simple_vocab_16e6.txt.gz (even if i agree my vocab file is 2MB heavier)

catwell: Yes, I know we do the same thing for CLIP. We use CLIP a lot so it simplified things initially. I am not sure we'd do the same thing today.

LouisRouss:
Hi,
Just realized that I didn't pass the CI/CD because of Pyright, I'm aware of the issue and will correct every Pyright errors this evening. Apologies for the setback

src/refiners/foundationals/fuyu/tokenizer.py
LouisRouss: I now create a .cache/refiners/fuyu-8b folder in the /home folder, save the tokenizer.json.gz there and retrieve it in the FuyuTokenizer class

src/refiners/foundationals/fuyu/transformers.py
```diff
@@ -0,0 +1,258 @@
...
+        super().__init__()
+        self.dim = dim
+        self.base = base
+        self.device = device
+        self.dtype = dtype
+        # Create positional encodings
+        self.theta = 1.0 / (self.base ** (torch.arange(0, self.dim, 2, dtype=torch.int64).float() / self.dim)).to(
+            self.device
+        )
+        self.cos: Tensor = torch.empty(0, device=self.device, dtype=self.dtype)
```
Laurent2916, line 29: nit
```suggestion
        self.theta = torch.pow(base, torch.arange(0, self.dim, 2, device=device, dtype=torch.float32) / self.dim).reciprocal()
```

src/refiners/foundationals/fuyu/transformers.py
```diff
@@ -0,0 +1,258 @@
...
+        self.sin: Tensor = torch.empty(0, device=self.device, dtype=self.dtype)
+
+    def _cache(self, seq_len: int) -> None:
+        if seq_len < self.cos.shape[0]:
+            return
+        t = torch.arange(seq_len, device=self.device, dtype=torch.int64).float()
+        freqs = torch.outer(t, self.theta)
```
Laurent2916, line 36: nit
```suggestion
        t = torch.arange(seq_len, device=self.device, dtype=torch.float32)
```

src/refiners/foundationals/fuyu/transformers.py
```diff
@@ -0,0 +1,258 @@
...
+                    fl.LayerNorm(normalized_shape=self.heads_dim, eps=self.norm_eps, device=device, dtype=dtype),
+                ),
+                fl.Chain(
+                    fl.Slicing(  # V projection
+                        dim=-2,
+                        start=2,
+                        end=3,
+                    ),
+                    fl.Squeeze(dim=-2),
+                ),
```
Laurent2916, line 136: nit fmt
```suggestion
            fl.Parallel(
                fl.Chain(  # Q projection
                    fl.Slicing(dim=-2, start=0, end=1),
                    fl.Squeeze(dim=-2),
                    fl.LayerNorm(normalized_shape=self.heads_dim, eps=self.norm_eps, device=device, dtype=dtype),
                ),
                fl.Chain(  # K projection
                    fl.Slicing(dim=-2, start=1, end=2),
                    fl.Squeeze(dim=-2),
                    fl.LayerNorm(normalized_shape=self.heads_dim, eps=self.norm_eps, device=device, dtype=dtype),
                ),
                fl.Chain(  # V projection
                    fl.Slicing(dim=-2, start=2, end=3),
                    fl.Squeeze(dim=-2),
                ),
```

src/refiners/foundationals/fuyu/transformers.py
```diff
@@ -0,0 +1,258 @@
...
+                embedding_dim=self.embedding_dim,
+                num_heads=self.num_heads,
+                use_bias=self.use_bias,
+                norm_eps=self.norm_eps,
+                device=device,
+                dtype=dtype,
+            ),
+            RotaryPositionalEmbedding(
+                dim=int(self.heads_dim * self.partial_rotary_factor), base=self.base, device=device, dtype=dtype
+            ),
```
Laurent2916, line 189: nit
```suggestion
        self.rotary_dim = int(self.heads_dim * self.partial_rotary_factor)

        super().__init__(
            QKVProjection(
                embedding_dim=self.embedding_dim,
                num_heads=self.num_heads,
                use_bias=self.use_bias,
                norm_eps=self.norm_eps,
                device=device,
                dtype=dtype,
            ),
            RotaryPositionalEmbedding(
                dim=self.rotary_dim,
                base=self.base,
                device=device,
                dtype=dtype,
```

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,81 @@
...
+    Warning:
+        The ref model from transformers can't be put on the device without the installation of accelerate
+
+    Raises:
+        AssertionError: If the outputs of the models differ by a margin greater than 1e-3.
+    """
+
+    manual_seed(42)
+    x = torch.rand(3, 512, 512)
+    x_pil: Image.Image = to_pil_image(x)  # type: ignore[reportUnknownVariableType]
```
Laurent2916, line 66: you could use refiners.fluxion.utils.tensor_to_image, and get rid of the reportUnknownVariableType error

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,81 @@
...
+    return processor
+
+
+@pytest.fixture(scope="module")
+def ref_model(test_device: Device) -> FuyuForCausalLM:
+    model: FuyuForCausalLM = FuyuForCausalLM.from_pretrained(pretrained_model_name_or_path="adept/fuyu-8b").to(  # type: ignore
+        device=test_device
+    )
+    return model  # type: ignore[reportUnknownVariableType]
+
```
Laurent2916, line 45: nit, you can probably simplify a bit the code here
```suggestion
@pytest.fixture(scope="module")
def ref_processor(test_device: torch.device) -> FuyuProcessor:
    return FuyuProcessor.from_pretrained(  # type: ignore
        pretrained_model_name_or_path="adept/fuyu-8b",
        device=test_device,
    )


@pytest.fixture(scope="module")
def ref_model(test_device: Device) -> FuyuForCausalLM:
    return FuyuForCausalLM.from_pretrained(  # type: ignore
        pretrained_model_name_or_path="adept/fuyu-8b",
        device=test_device,
    )

```

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,81 @@
...
+        "Describe this image. \n",
+        "Is there a cat in the image? \n",
+        "What is the emotion of the person? \n",
+        "What is the main object in this image? \n",
+    ]
+    p = random.choice(prompts)
+
+    with no_grad():
+        ref_input = ref_processor(text=p, images=x_pil, return_tensors="pt").to(device=test_device)  # type: ignore[reportUnknownMemberType]
+        ref_output = ref_model(**ref_input)["logits"]
```
Laurent2916, line 78: nit, is the `.to` required ?
```suggestion
        ref_input = ref_processor(text=p, images=x_pil, return_tensors="pt")
```

src/refiners/foundationals/fuyu/fuyu.py
```diff
@@ -0,0 +1,267 @@
...
+        num_heads=config.num_heads,
+        norm_eps=config.norm_eps,
+        base=config.base,
+        partial_rotary_factor=config.partial_rotary_factor,
+        use_bias=config.use_bias,
+        is_optimized=config.is_optimized,
+        device=config.device,
+        dtype=config.dtype,
+    )
+    return model
```
Laurent2916, line 81: no need to make config dataclasses, we'll simply use the constructor of the Fuyu class.
though you could declare a Fuyu8b class which is "pre-filled" with the above config, a bit like how it's done with DINOv2.

src/refiners/foundationals/fuyu/fuyu.py
```diff
@@ -0,0 +1,267 @@
...
+                for _ in range(num_layers)
+            ),
+            fl.LayerNorm(normalized_shape=embedding_dim, eps=norm_eps, device=device, dtype=dtype),
+            fl.Linear(in_features=embedding_dim, out_features=vocabulary_size, bias=False, device=device, dtype=dtype),
+        )
+
+    def init_context(self) -> dict[str, dict[str, Any]]:
+        return {"attention": {"mask": None}}
+
+    def generate(self, images: List[Image.Image], prompts: List[str], max_len_generation: int = 50):
```
Laurent2916, line 147: nit, please add a unit test, and maybe break down the function a bit

src/refiners/foundationals/fuyu/tokenizer.py
```diff
@@ -0,0 +1,210 @@
...
+    def __init__(
+        self,
+        vocabulary_path: str | Path | None = None,
+    ) -> None:
+        super().__init__()
+
+        if vocabulary_path is None:
+            vocabulary_path = Path.home() / ".cache/refiners/fuyu-8b/tokenizer.json.gz"
+            if not vocabulary_path.exists():
+                raise FileNotFoundError(f"Vocabulary file not found at {vocabulary_path}")
```
Laurent2916, line 37: nit, simply make the vocabulary_path required (i.e. no need to set it to None by default)

Laurent2916 (commented):
This looks pretty good, I made a first surface review, I haven't delved too deep inside the code logic (mostly for `input_processor.py` and `tokenizer.py`, as they look pretty beefy).

Note for other maintainers, disregard docstrings, the  majority of them look okay, I'll reword them after this PR to "harmonise" them

scripts/conversion/convert_transformers_fuyu8b.py
```diff
@@ -0,0 +1,181 @@
...
+    # create refiners .cache folder
+    refiners_cache_path = Path.home() / ".cache/refiners/fuyu-8b/"
+    refiners_cache_path.mkdir(parents=True, exist_ok=True)
+
+    model_id = "adept/fuyu-8b"
+
+    # extract the json vocabulary file from the hf .cache
+    processor: FuyuProcessor = FuyuProcessor.from_pretrained(pretrained_model_name_or_path=model_id)  # type: ignore[reportUnknownMemberType, reportAssignmentType]
+    vocabulary_file_path: str = Path(processor.tokenizer.vocab_file).absolute().parent / "tokenizer.json"  # type: ignore[reportUnknownMemberType, reportUnknownVariableType]
+    vocabulary_save_path = refiners_cache_path / "tokenizer.json.gz"
```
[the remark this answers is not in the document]
Laurent2916: no wait, we may want to simply ask the user for the Path to the tokenizer vocab file
cc @deltheil

src/refiners/foundationals/fuyu/common.py
```diff
@@ -0,0 +1,298 @@
...
+    This layer reshapes the input tensor to a specific shape (which must be compatible with the original shape).
+    See also [torch.reshape].
+
+    Warning:
+        The first dimension and seconde dimension (batch dimension and
+        sequence length) are forcefully preserved.
+
+    Example:
+        ```py
+        reshape = CustomReshape(5, 2)
```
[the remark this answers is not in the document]
limiteinductive: To add more context to @Laurent2916 reply, fl.Reshape always preserves the batch size to use -1 for another dimension.

tests/foundationals/fuyu/test_fuyu.py
LouisRouss: Indeed directly returning simplifies the code, however using the "device_map" argument for from_pretained requires accelerate. 

"Using `low_cpu_mem_usage=True` or a `device_map` requires Accelerate: `pip install accelerate"

tests/foundationals/fuyu/test_fuyu.py
LouisRouss: Unfortunately even with device_map set for FuyuProcessor, the ref_input still needs to be put manually to the test_device

tests/foundationals/fuyu/test_fuyu.py
Laurent2916: oh you're right, I missed that since I had accelerate installed.

src/refiners/foundationals/fuyu/fuyu.py
LouisRouss: Created a pre-filled Fuyu8b class in which tokenizer_path, device and dtype need to be specified.

src/refiners/foundationals/fuyu/tokenizer.py
LouisRouss: The vocabulary path is now required

src/refiners/foundationals/fuyu/common.py
```diff
@@ -0,0 +1,298 @@
...
+        See also `merge_multi_head`, which is the inverse operation.
+        """
+        assert (
+            x.ndim == 3
+        ), f"Expected input tensor with shape (batch_size sequence_length embedding_dim), got {x.shape}"
+        assert (
+            x.shape[-1] % self.num_heads == 0
+        ), f"Expected embedding_dim (x.shape[-1]={x.shape[-1]}) to be divisible by num_heads ({self.num_heads})"
+
+        return x.reshape(x.shape[0], x.shape[1], self.num_heads, x.shape[-1] // self.num_heads).transpose(1, 2)
```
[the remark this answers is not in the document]
LouisRouss: Removed duplicated code with inheritance of fl.ScaledDotProductAttention

src/refiners/foundationals/fuyu/common.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def _merge_multi_head(
+        self,
+        x: Float[Tensor, "batch_size num_heads sequence_length heads_dim"],
+    ) -> Float[Tensor, "batch_size sequence_length heads_dim * num_heads"]:
+        """Merge the input tensor from multiple heads along the embedding dimension.
+
+        See also `split_to_multi_head`, which is the inverse operation.
+        """
+        return x.transpose(1, 2).reshape(x.shape[0], x.shape[2], self.num_heads * x.shape[-1])
```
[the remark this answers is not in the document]
LouisRouss: Removed duplicated code with inheritance of fl.ScaledDotProductAttention

src/refiners/foundationals/fuyu/__init__.py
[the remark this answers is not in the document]
LouisRouss: I now export Fuyu and Fuyu8b

src/refiners/foundationals/fuyu/transformers.py
LouisRouss: indeed, expression simplified !

src/refiners/foundationals/fuyu/transformers.py
LouisRouss: indeed, expression simplified !

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Warning:
+        The ref model from transformers can't be put on the device without the installation of accelerate
+
+    Raises:
+        AssertionError: If the outputs of the models differ by a margin greater than 1e-3.
+    """
+
+    manual_seed(42)
+    x = (torch.randint(0, 256, size=(1, 3, 512, 512), dtype=torch.uint8) / 255).float()
```
LouisRouss, line 63: Need to do change the fixed resolution as done in dinov2

LouisRouss, line 63: I changed 
```python
x = torch.rand(1,3,512,512)
```
to 
```python
x = (torch.randint(0, 256, size=(1, 3, 512, 512), dtype=torch.uint8) / 255).float()
```
I faced an issue with the first expression illustrated in the following screenshot 
image

I must confess i don't have much idea on the reason of the problem, if you have any hint I would gladly welcome it.

src/refiners/foundationals/fuyu/common.py
```diff
@@ -0,0 +1,298 @@
...
+        # the attention is upcast to float32
+        dtype = query.dtype
+        return self._merge_multi_head(
+            x=self.dot_product(
+                query=self._split_to_multi_head(query.float()),
+                key=self._split_to_multi_head(key.float()),
+                value=self._split_to_multi_head(value.float()),
+                attn_mask=attn_mask,
+            ).to(dtype)
+        )
```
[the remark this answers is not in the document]
LouisRouss: In the reference implementation the upcast is done only for the softmax part : 
https://github.com/huggingface/transformers/blob/fbb41cd4209c00671fef583a39db2f608fcd7098/src/transformers/models/persimmon/modeling_persimmon.py#L338

In both cases (the whole dot product upcast vs only softmax) the model doesn't pass the max absolute difference between output logits test in float16 (in float32 the difference is 0). I'm open to any feedback on this. 

Even if the generated text remain mostly consistent between float32 and float16 with the softmax upcast, I found some cases where the text generated in float16 deviate a little from the one generated in float32. Upcasting the whole dot product part to float32 seemed to solve the problem (see example screenshot below). I'm aware that it remains some isolated empirical examples and no real conclusion can be drawn.

image: "https://huggingface.co/adept/fuyu-8b/resolve/main/bus.png"
prompt : "Generate a caption for this image. \n"

Refiners float16 whole attn upcast 
image

Refiners float16 only softmax upcast
image

HuggingFace float32
image

scripts/conversion/convert_transformers_fuyu8b.py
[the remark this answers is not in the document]
LouisRouss: I removed the dependency to transformers, the user needs to specify an output path where the vocabulary file and the weights will be stored

scripts/conversion/convert_transformers_fuyu8b.py
```diff
@@ -0,0 +1,181 @@
...
+    # compress the json vocabulary file
+    with open(vocabulary_file_path, "rb") as original_file:
+        with gzip.open(vocabulary_save_path, "wb") as compressed_file:
+            shutil.copyfileobj(original_file, compressed_file)
+
+    # get hf fuyu weights
+    source: FuyuForCausalLM = FuyuForCausalLM.from_pretrained(pretrained_model_name_or_path=model_id)  # type: ignore[reportUnknownMemberType, reportAssignmentType]
+    weights = source.state_dict()
+    del source
+    gc.collect()
```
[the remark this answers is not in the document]
LouisRouss: I removed the dependency and now use hf_hub_download

src/refiners/foundationals/fuyu/fuyu.py
LouisRouss: unit test added + function broken down

src/refiners/foundationals/fuyu/fuyu.py
```diff
@@ -0,0 +1,296 @@
...
+        Receives:
+            images (List[Image.Image, "batch"])
+            prompts (List[str, "batch"])
+            max_len_generation (int)
+
+        Returns:
+            (List[str, "batch"])
+        """
+        tensor_images = [(Tensor(np.array(image) / 255)).permute(2, 0, 1).unsqueeze(0) for image in images]
+
```
Laurent2916, line 110: nit
```suggestion
        tensor_images = images_to_tensor(images)
```

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,130 @@
...
+from refiners.fluxion.utils import load_from_safetensors, manual_seed, no_grad, tensor_to_image
+from refiners.foundationals.fuyu.fuyu import Fuyu, Fuyu8b
+
+
+def _img_open(path: Path) -> Image.Image:
+    return Image.open(path).convert("RGB")  # type: ignore
+
+
+@pytest.fixture(scope="module")
+def our_model(test_weights_path: Path, test_vocab_path: Path, test_device: Device) -> Fuyu:
```
Laurent2916, line 21: the `test_vocab_path` is missing

Laurent2916 (commented):
Thanks for applying the suggestions, the code looks alright, though I feel it could be a bit more "fluxionny", wdyt @deltheil @limiteinductive ?
Also, I was able to run the models (in float16) and I confirmed that this implementation's output was close to the huggingface's implementation.

tests/foundationals/fuyu/test_fuyu.py
```diff
@@ -0,0 +1,130 @@
+import random
+from pathlib import Path
+from typing import List
```
limiteinductive, line 3: ditto

src/refiners/foundationals/fuyu/input_processor.py
```diff
@@ -0,0 +1,275 @@
+from typing import List, Tuple
```
limiteinductive, line 1: ditto

limiteinductive (changes requested):
Thanks for the great work. I haven't been able to run the tests; I must use a VM with more VRAM. It would be more practical to store some outputs and compare your implementation to them instead of loading both models simultaneously.

Regarding @Laurent2916's remark about making it more "fluxiony"—which means making the most of Fluxion's capability and writing as few forwards as possible—there are probably some places where we could improve things. Still, we can always do that later when we need it.

src/refiners/foundationals/fuyu/fuyu.py
```diff
@@ -0,0 +1,296 @@
+import re
+from pathlib import Path
+from typing import Any, List
```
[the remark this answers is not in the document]
LouisRouss: List is replaced by list

src/refiners/foundationals/fuyu/fuyu.py
```diff
@@ -0,0 +1,296 @@
...
+
+        Receives:
+            images (List[Image.Image, "batch"])
+            prompts (List[str, "batch"])
+            max_len_generation (int)
+
+        Returns:
+            (List[str, "batch"])
+        """
+        tensor_images = [(Tensor(np.array(image) / 255)).permute(2, 0, 1).unsqueeze(0) for image in images]
```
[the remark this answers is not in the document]
LouisRouss: images_to_tensor and tensor_to_images are now used

src/refiners/foundationals/fuyu/tokenizer.py
```diff
@@ -0,0 +1,205 @@
...
+                self.token_bbox_open,
+                self.token_bbox_close,
+                self.token_point_open,
+                self.token_point_close,
+            ]:
+                continue
+            elif i > 0 and text_split[i - 1] in [self.token_bbox_open, self.token_point_open]:
+                tokens += self.process_points_coordinates(elem, scale_factor=scale_factor)
+            else:
+                tokens += self.process_text(elem)
```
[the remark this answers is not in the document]
LouisRouss: Changed to a match/case syntax

src/refiners/foundationals/fuyu/input_processor.py
```diff
@@ -0,0 +1,275 @@
...
+    ) -> None:
+        self.patch_size = patch_size
+        self.embedding_dim = embedding_dim
+        self.padding_value = padding_value
+        self.image_mean = image_mean
+        self.image_std = image_std
+
+        super().__init__(
+            PatchPadding(patch_size=self.patch_size, padding_value=self.padding_value),
+            fl.Lambda(func=self.normalize),
```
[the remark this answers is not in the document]
LouisRouss: Oh indeed
done !

src/refiners/foundationals/fuyu/fuyu.py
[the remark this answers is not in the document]
LouisRouss: I now use 
```python
tensor_images = [image_to_tensor(image) for image in images]
``` 
I can't use `refiners.fluxion.utils.images_to_tensor` as I keep all images at their original resolution during the forward pass

src/refiners/foundationals/fuyu/transformers.py
```diff
@@ -0,0 +1,250 @@
...
+        self.dtype = dtype
+        # Create positional encodings
+        self.theta = torch.pow(
+            base, torch.arange(0, self.dim, 2, device=device, dtype=torch.float32) / self.dim
+        ).reciprocal()
+
+        self.cos: Tensor = torch.empty(0, device=self.device, dtype=self.dtype)
+        self.sin: Tensor = torch.empty(0, device=self.device, dtype=self.dtype)
+
+    def _cache(self, seq_len: int) -> None:
```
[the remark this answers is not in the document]
LouisRouss: The method is designed to precompute and store the cosines and sines only recalculating them if a longer sequence is encountered across the forward passes. The term cache might be a slight misnomer, I can change the method name

tests/foundationals/fuyu/test_fuyu.py
LouisRouss: I added the fixture, assuming having a "test_vocab" folder in the test fuyu folder. Perhaps you would prefer having something more generic added to the conftest.py ?

LouisRouss:
Thank you for the reviews ! Concerning the test script, i know generate the references in a conftest.py before the different test_ functions. I hope this solution will be satisfying :)

Laurent2916:
Hi @LouisRouss, sorry for the delay!
we're very busy with other projects at the moment, I will keep the PR open

PR #304, opened 2024-02-23
BaseBatch : easily manipulate batched list & tensors

piercus:
This is a proposal for a `Batch` class to be used when training.

## Context

When training, we always need to manipulate batched data, those data are generally Tensors (where batch_size is `shape[0]`) and lists (where batch_size is `len(lst)`)

Morever, it is quite classic to get mixed up between `list[Image.Image]` or `Tensor` that might represent the same thing (a group of images), so having the flexbility to switch the structure of the batches easily in one line while keeping the rest of the batching logic safe is quite convenient.

## Proposal

This `BaseBatch` class will both provide the ability to handle this, while providing type checks and size checks at different levels to raise the error as soon as possible in the process.

### API

```python
from refiners.training_utils.batch import BaseBatch
from torch import Tensor

class MockBatch(BaseBatch):
    foo: Tensor
    bar: Tensor
    indices: list[int]
```
### Features

* Checking that hint are list or tensor when inherinting
* Making sure all the args have the same (non-zero) size
* Even if you do not use pyright, first-level-typing is checked when you inherit from `BaseBatch`
* Training related features : `cls.collate`, `inst.to`, `cls.load`, `inst.save`
* Pythonic stuff like `__len__`, `__iter__`, `to_dict`, `clone`, `__eq__`, `__getitem__`, `__repr__`

### Questions

* Currently Batch cannot have 0-size, this can be discussed.
* The `batch.py` and `test_batch.py` could also have been put in `training.py` and `test_training.py`, please share your thoughts
* Not sure between `BaseBatch`, `AbstractBatch` or simply `Batch` as the top-level name

## Going further

There might be a second level PR for this, with jax-typed batch, but i feel it can be done incrementally, and it's easier to discuss this step by step.

### DIFF
--- a/src/refiners/training_utils/batch.py
+++ b/src/refiners/training_utils/batch.py
@@ -1,7 +1,9 @@
 from pathlib import Path
-from typing import Any, Type, TypeVar, get_origin, get_type_hints
+from typing import Any, Type, TypeVar, cast, get_origin, get_type_hints
 
 from torch import Tensor, cat, device as Device, dtype as DType, load as torch_load, save as torch_save  # type: ignore
+
+from refiners.fluxion.utils import summarize_tensor  # type: ignore
 
 T = TypeVar("T", bound="BaseBatch")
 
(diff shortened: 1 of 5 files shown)


src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def __repr__(self) -> str:
+        attr_strs: list[str] = []
+        for type_key in get_type_hints(self.__class__):
+            attr_value = getattr(self, type_key)
+            if isinstance(attr_value, list):
+                attr_strs.append(f"{type_key}={attr_value}")
+            else:
+                attr_strs.append(f"{type_key}={summarize_tensor(attr_value)}")
+        return f"{self.__class__.__name__}({len(self)})[{','.join(attr_strs)}]"
```
limiteinductive, line 172: Let's maybe be explicit about what is in the repr
```suggestion
        return f"{self.__class__.__name__}(size={len(self)})[{','.join(attr_strs)}]"
```

src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,194 @@
...
+
+            if isinstance(attr_list[0], Tensor):
+                collated_attrs[type_key] = cat(tensors=tuple(attr_list), dim=0)
+            elif isinstance(attr_list[0], list):
+                collated_attrs[type_key] = [item for sublist in attr_list for item in sublist]
+            else:
+                raise ValueError(f"Unsupported attribute type for collation: {type(attr_list[0])}")
+
+        collated_instance = cls(**collated_attrs)
+        return collated_instance
```
limiteinductive, line 103:
```suggestion

    def __getattr__(self, name: str) -> AttrType:
        if name in get_type_hints(self.__class__):
            return getattr(self, name)
        else:
            raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
            
    @classmethod
    def collate(cls: Type[T], batch_list: list[T]) -> T:
        collated_attrs: dict[str, Any] = {}
        type_hints = get_type_hints(cls)

        if len(batch_list) == 0:
            raise ValueError(f"Cannot collate an empty list of {cls.__name__}")

        for type_key in type_hints.keys():
            attr_list: list[Tensor | list[Any]] = [obj.__getattr__(type_key) for obj in batch_list]

            if all(isinstance(attr, Tensor) for attr in attr_list):
                tensor_tuple = cast(tuple[Tensor, ...], tuple(attr_list))
                collated_attrs[type_key] = cat(tensor_tuple, dim=0)
            else:
                collated_attrs[type_key] = [item for sublist in attr_list for item in sublist]

        collated_instance = cls(**collated_attrs)
        return collated_instance
```

limiteinductive: You can probably also make a __setattr__ magic method

limiteinductive (changes requested):
An issue is that you don't get a good signature when subclassing BaseBatch

image

piercus:
@limiteinductive thanks for your help.

I've used [`@dataclass_transform()`](https://peps.python.org/pep-0681/) to fix the typing signature.

Also added `__add`, `__setattr__`

src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,222 @@
...
+            attr_type = attr_types[name]
+            if not isinstance(value, attr_type):
+                raise TypeError(
+                    f"Invalid type for attribute '{name}': Expected '{attr_type.__name__}', got '{type(value).__name__}'"
+                )
+            if isinstance(value, list):
+                new_size = len(value)
+            else:
+                new_size = value.shape[0]
+
```
limiteinductive, line 86:
```suggestion
            new_size = len(value) if isinstance(value, list) else value.shape[0]
```

src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,222 @@
...
+    def collate(cls: Type[T], batch_list: list[T]) -> T:
+        collated_attrs: dict[str, Any] = {}
+        attr_types = cls.attr_types()
+
+        if len(batch_list) == 0:
+            raise ValueError(f"Cannot collate an empty list of {cls.__name__}")
+
+        for attr_name, attr_type in attr_types.items():
+            attr_list = [getattr(obj, attr_name) for obj in batch_list]
+
```
limiteinductive, line 104: Otherwise you don't get correct typing
```suggestion
            attr_list = [obj.__getattr__(attr_name) for obj in batch_list]
```

src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,222 @@
...
+    def to(self: T, device: Device | None = None, dtype: DType | None = None) -> T:
+        attr_types = self.__class__.attr_types()
+        for attr_name, attr_type in attr_types.items():
+            if attr_type == Tensor:
+                value = cast(Tensor, self.__getattr__(attr_name))
+                self.__setattr__(attr_name, value.to(device, dtype))
+
+        return self
+
+    def attr_length(self, name: str) -> int:
```
limiteinductive, line 137: shouldn't this be attr_size?

src/refiners/training_utils/batch.py
```diff
@@ -0,0 +1,222 @@
...
+        return self.attr_length(list(self.__class__.attr_types().keys())[0])
+
+    def to_dict(self) -> dict[str, AttrType]:
+        return {attr_name: getattr(self, attr_name) for attr_name in self.__class__.attr_types()}
+
+    def split(self: T) -> list[T]:
+        result: list[T] = []
+        l = len(self)
+        for i in range(l):
+            args = {attr_name: getattr(self, attr_name)[i : i + 1] for attr_name in self.__class__.attr_types()}
```
limiteinductive, line 154:
```suggestion
        for i in range(len(self)):
```

limiteinductive (approved):
@deltheil PTAL

PR #321, opened 2024-03-13
Make LoRA's weight initialization overridable

Laurent2916:
As per Section 4.1 of the paper (https://arxiv.org/abs/2106.09685), we initialize the weights of the LoRA like so:
- up -> zeros
- down -> normal (gaussian)

Though in pratice, many other implementation do the following:
- up -> zeros
- down -> kaiming_uniform

see Microsoft's code: https://github.com/microsoft/LoRA/blob/4c0333854cb905966f8cc4e9a74068c1e507c7b7/loralib/layers.py#L122-L125

Let's make the code slightly more generic by allowing the initialization to be overridable.


deltheil (dismissed):

deltheil (approved):
@Laurent2916 perhaps you want to tweak the PR description a little bit more

PR #326, opened 2024-03-19
Switch gradient clipping to native torch torch.nn.utils.clip_grad_norm_

limiteinductive:
Currently, we have a custom function to compute and clip grad_norm. We switch to using a `torch.nn.utils.clip_grad_norm_` with internal optimizations to compute multiple vectors in parallel.

Also, we switch from clipping after each backward step to clipping before each optimization step, following the convention set by PyTorch. This is not equivalent to the previous behavior when `gradient_accumulation > 1`. The `WandbCallback` has been updated to reflect that change and will log the `total_grad_norm`after an optimizer step.


src/refiners/training_utils/common.py
```diff
@@ -7,19 +7,18 @@
...
 
 from refiners.fluxion.utils import manual_seed
 
 
 def compute_grad_norm(parameters: Iterable[nn.Parameter]) -> float:
     """
-    Computes the gradient norm of the parameters of a given model similar to `clip_grad_norm_` returned value.
+    Computes the gradient norm of the parameters in the given iterable.
+
+    We use the `torch.nn.utils.clip_grad_norm_` function to compute the gradient efficiently on the GPU or CPU.
```
deltheil, line 19:
```suggestion
    We use the `torch.nn.utils.clip_grad_norm_` function to process the gradients efficiently on the GPU or CPU.
```

deltheil (dismissed):

deltheil (approved):

PR #344, opened 2024-04-02
Fix typos ignores

Laurent2916:
[`typos`](https://github.com/crate-ci/typos/) github action was updated to v1.20.1.
It now contains the rule [`arange->arrange`](https://github.com/crate-ci/typos/issues/955) which conflicts with every `torch.arange` call.

I also added a `.ignore` file which tells `typos` which files/directories to ignore, most notably we want to ignore `test/repos`, as third party repos may contains lots of typos.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,8 +143,11 @@ exclude_also = [
 ]
 
 [tool.typos.default]
-extend-words = { adaptee = "adaptee" }
 extend-ignore-identifiers-re = ["NDArray*", "interm", "af000ded"]
+
+[tool.typos.default.extend-words]
+adaptee = "adaptee"  # Common name for an adapter's target
+arange = "arange"  # Commonly used via `from torch import arange`
 
 [tool.pytest.ini_options]
 filterwarnings = [


.ignore
deltheil, line 1: Can we add a prefix like `.typos.ignore`? Otherwise, can you add a comment on top like `# For typos (spell checker`

Laurent2916: Well it looks like `typos` ignores by default the files/directories inside `.gitignore`, so we actually don't need the `.ignore` file. I was on a branch where `test/repos` wasn't in the `.gitignore`, my bad.

pyproject.toml
```diff
@@ -143,7 +143,7 @@ exclude_also = [
 ]
 
 [tool.typos.default]
-extend-words = { adaptee = "adaptee" }
+extend-words = { adaptee = "adaptee", "arange" = "arange" }
```
deltheil, line 146: I would drop the quotes for the key for consistency (`arange = "arange"`) or use this variant for extra inline comment:

```toml
[tool.typos.default]
extend-ignore-identifiers-re = ["NDArray*", "interm", "af000ded"]

[tool.typos.default.extend-words]
adaptee = "adaptee"
# Commonly used via `from torch import arange`
arange = "arange"
```

deltheil (approved):

pyproject.toml
Laurent2916: good idea, let's do that

deltheil (approved):

PR #345, opened 2024-04-02
update pyright and fix Pillow 10.3 typing issues

catwell:


tests/e2e/test_diffusion.py
```diff
@@ -34,6 +34,10 @@
 from tests.utils import ensure_similar_images
 
 
+def _img_open(pth: Path) -> Image.Image:
+    return Image.open(pth)  # type: ignore
```
Laurent2916, line 38: Should we move this inside `refiners.fluxion.utils` ?

tests/e2e/test_diffusion.py
```diff
@@ -34,6 +34,10 @@
 from tests.utils import ensure_similar_images
 
 
+def _img_open(pth: Path) -> Image.Image:
```
deltheil, line 37: Is `pth` intended (short form) or you just wanted to use `path` instead?

tests/e2e/test_doc_examples.py
```diff
@@ -13,6 +13,10 @@
 from tests.utils import ensure_similar_images
 
 
+def _img_open(pth: Path) -> Image.Image:
```
deltheil, line 16: ditto

deltheil (dismissed):

tests/e2e/test_diffusion.py
catwell: Hmm, intended, I tend to name paths `pth` (to avoid conflicts with `from os import path` mostly). I can rename it if you prefer.

deltheil: Given that we have `path: ...,` elsewhere I'd say. But no big deal.

deltheil (approved):

PR #367, opened 2024-04-23
Fix `StableDiffusion_1_Inpainting` init

Laurent2916:
Currently, if not provided, the `SD1Unet` of an `StableDiffusion_1_Inpainting` will be initialized with 4 channels instead of 9 (because of inheritance with `StableDiffusion_1`)

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
+++ b/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
@@ -198,8 +198,14 @@ class StableDiffusion_1_Inpainting(StableDiffusion_1):
     ) -> None:
         self.mask_latents: Tensor | None = None
         self.target_image_latents: Tensor | None = None
+        unet = unet or SD1UNet(in_channels=9)
         super().__init__(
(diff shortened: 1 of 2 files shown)


src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
```diff
@@ -198,8 +198,14 @@ def __init__(
     ) -> None:
         self.mask_latents: Tensor | None = None
         self.target_image_latents: Tensor | None = None
+        unet = unet or SD1UNet(in_channels=9)
```
limiteinductive, line 201: You should pass the device and dtype to the constructor

limiteinductive (changes requested):
See small remark otherwise lgtm

src/refiners/foundationals/latent_diffusion/stable_diffusion_1/model.py
deltheil, line 1: Adding an unit test would not hurt (e.g. under `test_model.py`?)

deltheil (approved):

limiteinductive (approved):

PR #402, opened 2024-08-20
Remove some `from torch import ...` from the code

Laurent2916:
See https://huggingface.co/spaces/zero-gpu-explorers/README/discussions/85
I also find it better to have the `torch.` prefix

I ran all the tests just in case (with `torch<2.4.0`), everything is good (~~I fixed the ic-light test, it got broken in a previous commit because we switched a default~~ -> https://github.com/finegrain-ai/refiners/pull/403)

### DIFF
--- a/src/refiners/fluxion/utils.py
+++ b/src/refiners/fluxion/utils.py
@@ -8,15 +8,7 @@ from numpy import array, float32
 from PIL import Image
 from safetensors import safe_open as _safe_open  # type: ignore
 from safetensors.torch import save_file as _save_file  # type: ignore
-from torch import (
-    Tensor,
-    cat,
-    device as Device,
-    dtype as DType,
-    manual_seed as _manual_seed,  # type: ignore
-    no_grad as _no_grad,  # type: ignore
-    norm as _norm,  # type: ignore
-)
+from torch import Tensor, device as Device, dtype as DType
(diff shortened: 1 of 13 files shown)


src/refiners/fluxion/utils.py
```diff
@@ -8,30 +8,22 @@
...
-from torch import (
-    Tensor,
-    cat,
-    device as Device,
-    dtype as DType,
-    manual_seed as _manual_seed,  # type: ignore
-    no_grad as _no_grad,  # type: ignore
-    norm as _norm,  # type: ignore
-)
+from torch import Tensor, device as Device, dtype as DType
```
deltheil, line 11: Why are some device as Device (and dtype as DType) are replaced (torch.device, ...) and some others kept as-is? No way to be more consistent?

Laurent2916: :sweat_smile: I began to replace every `from torch import ...`, but saw that it would produce a very large diff (because of the `device as Device`, `dtype as DType`, ...), so I changed my mind and only modified the other imports like `cat`, `stack`, `zeros`, ... I completely forgot to revert the extra changes, let me fix this (I'll make a follow-up PR for these remnant imports to make everything more consistent globaly)
I'll also split the ic-light test fix in an other PR

Laurent2916: new PR for the IC-Light fix: #403

deltheil (approved):

PR #411, opened 2024-08-26
Add Multi-View Aggregation Network (MVANet)

catwell:


src/refiners/foundationals/swin/mvanet/converter.py
```diff
@@ -0,0 +1,138 @@
...
+
+def convert_weights(official_state_dict: dict[str, Tensor]) -> dict[str, Tensor]:
+    rm_list = [
+        # Official weights contains useless keys
+        # See https://github.com/qianyu-dlut/MVANet/issues/3#issuecomment-2105650425
+        r"multifieldcrossatt.linear[56]",
+        r"multifieldcrossatt.attention.5",
+        r"dec_blk\d+\.linear[12]",
+        r"dec_blk[1234]\.attention\.[4567]",
+        # We don't need the sideout weights
```
deltheil, line 14: I guess this is "... at inference-time"

catwell: Yes, IMO if we add things just for training it must be in training-specific code (using adapters etc).

deltheil: Worth expanding to "We don't need the sideout weights at inference-time" then (just to make it 100% clear)?

src/refiners/foundationals/swin/mvanet/mcrm.py
```diff
@@ -0,0 +1,119 @@
...
+
+import torch
+from torch import Tensor, device as Device
+
+import refiners.fluxion.layers as fl
+
+from .utils import FeedForward, Interpolate, MultiPool, PatchMerge, PatchSplit, PatchwiseCrossAttention, Unflatten
+
+
+class Multiply(fl.Chain):
```
deltheil, line 11: Not to be confused with `src/refiners/fluxion/layers/basics.py`'s `Multiply` (nit: could be worth picking a different name - or just tackle later if/when we move it under fluxion utils)

deltheil (approved):

src/refiners/foundationals/swin/mvanet/mclm.py
```diff
@@ -0,0 +1,211 @@
...
+            pool_ratios = [2, 8, 16]
+
+        positional_embedding = PositionEmbeddingSine(num_pos_feats=emb_dim // 2, device=device)
+
+        # LayerNorms in MCLM share their weights.
+
+        ln1 = fl.LayerNorm(emb_dim, device=device)
+        ln2 = fl.LayerNorm(emb_dim, device=device)
+
+        def proxy(m: fl.Module) -> fl.Module:
```
deltheil, line 139: I was tempted to add another comment (or expand the above one re: layer norms shared weights), just to make it clear which role `proxy` has there.

src/refiners/foundationals/swin/mvanet/mcrm.py
deltheil: > I think we should keep it like this for now and discuss the name if we decide to put them in utils.

Agreed.

deltheil (approved):

PR #414, opened 2024-08-28
fix deprecation warning

catwell:
`DeprecationWarning: The 'warn' function is deprecated, use 'warning' instead`

### DIFF
--- a/src/refiners/training_utils/config.py
+++ b/src/refiners/training_utils/config.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from logging import warn
+from logging import warning
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/refiners/training_utils/config.py
```diff
@@ -1,5 +1,5 @@
 from enum import Enum
-from logging import warn
```
deltheil, line 2: We have many more throughout the tests: do you also want to change these ones?

deltheil: My bad, these are from the `warnings` package (not logging)

deltheil (approved):

PR #423, opened 2024-09-05
Add comfyui-refiners requirements.txt

Laurent2916:
apparently the deps declared in the pyproject.toml aren't enough


src/comfyui-refiners/requirements.txt
deltheil, line 1: Wouldn't `-e file:.` be enough? Like https://github.com/finegrain-ai/refiners/blob/main/requirements.docs.txt

Laurent2916: ah good idea, let's try that

deltheil (approved):

PR #427, opened 2024-09-06
MultiUpscaler: improve resizing logic

catwell:
- Do not have a parameter to downscale (caller can do it beforehand if they want).
- Do not enforce mod 8 *before* upscale, we need it afterwards only (before SD).

### DIFF
--- a/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/multi_upscaler.py
+++ b/src/refiners/foundationals/latent_diffusion/stable_diffusion_1/multi_upscaler.py
@@ -171,31 +171,18 @@ class MultiUpscalerAbstract(MultiDiffusion[T], ABC):
             condition_scale=target.condition_scale,
         )
 
-    @staticmethod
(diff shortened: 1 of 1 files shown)


src/refiners/foundationals/latent_diffusion/stable_diffusion_1/multi_upscaler.py
```diff
@@ -171,31 +171,19 @@ def diffuse_upscaler_target(self, x: Tensor, step: int, target: UpscalerTarget)
...
-
-        The `on_short` parameter determines whether the resizing is based on the shortest side.
-        """
-        assert size % 8 == 0, "Size must be a multiple of 8 because this is the latent compression size."
-        side_size = min(image.size) if on_short else max(image.size)
-        scale = size / (side_size * 8)
-        new_size = (int(image.width * scale) * 8, int(image.height * scale) * 8)
-        return image.resize(new_size, resample=Image.Resampling.LANCZOS)  # type: ignore
-
     @no_grad()
```
deltheil, line 174: Let's get rid of it?

deltheil (approved):

PR #428, opened 2024-09-08
Fix SD1UNet's UpBlocks missing device/dtype + expose SD1Autoencoder and SDXLAutoencoder

Laurent2916:
Somehow the dtype and device was missing for the UpBlocks.
Exposing SD1Autoencoder and SDXLAutoencoder directly in `refiners.foundationals.latent_diffusion` makes imports slightly easier


src/refiners/foundationals/latent_diffusion/stable_diffusion_1/unet.py
```diff
@@ -188,7 +188,7 @@ def __init__(
                 fl.UseContext(context="unet", key="residuals").compose(lambda x: x[-1]),
                 MiddleBlock(device=device, dtype=dtype),
             ),
-            UpBlocks(),
+            UpBlocks(device=device, dtype=dtype),
```
deltheil, line 191: Nice catch. Could you please add a test (for coverage) under `tests/foundationals/latent_diffusion/test_unet.py`?

(I guess in practical cases, we benefited from `self.unet = unet.to(device=self.device, dtype=self.dtype)` in `LatentDiffusionModel` CTOR)

Laurent2916: yes indeed the `LatentDiffusionModel`'s init fixed the problem, this problem arises when testing/using the bare unet. I'll add a simple "forward test".

deltheil (approved):

PR #442, opened 2024-09-19
Update installation instructions in README.md and CONTRIBUTING.md

Laurent2916:


README.md
```diff
@@ -56,12 +56,10 @@ ______________________________________________________________________
 
 ## Installation
 
-The current recommended way to install Refiners is from source using [Rye](https://rye-up.com/):
+The current recommended way to install Refiners is from source:
 
 ```bash
-git clone "git@github.com:finegrain-ai/refiners.git"
```
deltheil, line 62: I would keep this. And just add:

Alternatively:

```bash
pip install git+https://github.com/finegrain-ai/refiners.git
```

deltheil: @Laurent2916 wdyt?

Laurent2916: I feel like the whole "rye setup" is more "dev oriented" than "user oriented".
wdyt about simply adding a reference to the CONTRIBUTING.md if people want to setup a dev env ?

deltheil: True. But https://refine.rs/getting-started/recommended/ is rye-centric.

deltheil: Anyway, let's rebase and use one or the other way.

catwell: The recommended option on the website is Rye because you needed to convert the weights anyway. We could revisit this.

deltheil (approved):

README.md
Laurent2916: Ah right, there is still the weight conversion. I agree we should probably revisit this before modifying the README.

PR #448, opened 2024-09-25
fix DPM-Solver with bfloat16

catwell:


src/refiners/foundationals/latent_diffusion/solvers/dpm.py
```diff
@@ -121,15 +121,6 @@ def _generate_timesteps(self) -> torch.Tensor:
         np_space = np.linspace(offset, max_timestep, self.num_inference_steps + 1).round().astype(int)[1:]
         return torch.tensor(np_space).flip(0)
 
-    def _generate_sigmas(self) -> tuple[torch.Tensor, torch.Tensor]:
```
deltheil, line 124: OK, so this one was redundant (useless)

deltheil (approved):

src/refiners/foundationals/latent_diffusion/solvers/dpm.py
catwell: yes

PR #462, opened 2024-10-09
Rework weight conversions

Laurent2916:
That's a pretty fat PR:
- Deleted all old conversion scripts.
- Created a new `conversion` module which converts weight state_dicts based on simple mappings (most of the time), [inspired by imaginAIry](https://github.com/brycedrennan/imaginAIry/tree/49f2c25b6b9a1d753829e2049d2a8058f985925e/imaginairy/weight_management/weight_maps).
  - Mapping conversion is way more lighter/faster than using the old ModelConverter (this allows weights to be converted on the fly if necessary/wanted).
  - Each conversion is now declarative. The hash (and sometime git commit revision) of both the original weights and converted weights are now strictly checked (to avoid unwanted regressions).
- Created new test fixture `use_local_weights` that allows to run the tests either with local converted weights, or the ones from the [refiners huggingface org](https://huggingface.co/refiners).
- Modified all existing tests to use these new fixtures.
- Added DINOv2 similarity when testing images in e2e tests.
- Loosen some tests to be good on both a 3090 and an L4.

See the updated `CONTRIBUTING.md` for instructions.


catwell (approved):

catwell (approved):

tests/utils.py
```diff
@@ -1,25 +1,56 @@
...
 
-def compare_images(img_1: Image.Image, img_2: Image.Image) -> tuple[int, float]:
-    x1, x2 = (
-        torch.tensor(np.array(x).astype(np.float32)).permute(2, 0, 1).unsqueeze(0) / 255.0 for x in (img_1, img_2)
+dinov2_small_model = DINOv2_small()
+dinov2_small_model.load_from_safetensors(
+    dinov2.small.converted.local_path
+    if dinov2.small.converted.local_path.exists()
+    else dinov2.small.converted.hf_cache_path
+)
```
catwell, line 19: Don't do that (load a model when requiring a very frequently requested helper). Use a function + functools cache.

catwell (changes requested):

tests/utils.py
Laurent2916: ooops, right my bad

catwell (approved):

PR #473, opened 2024-10-23
latent_diffusion.solver: assert msg must be f-strings

piercus:
Solver step assert msg need to be f-strings


deltheil (approved):

catwell:
:facepalm: 

I thought the linter checked this but it is actually [in preview](https://docs.astral.sh/ruff/rules/missing-f-string-syntax/).

PR #474, opened 2024-11-06
make `refiners.conversion.utils.Hub.expected_sha256` optional

Laurent2916:
This way you can use a `Hub` object without specifying the hash, useful when declaring where to find a weight on the hub (since there is no need to check the hash).

### DIFF
--- a/src/refiners/conversion/utils.py
+++ b/src/refiners/conversion/utils.py
@@ -26,9 +26,10 @@ def download_file_url(url: str, destination: Path) -> None:
     logging.debug(f"Downloading {url} to {destination}")
 
     # get the size of the file
-    response = requests.get(url, stream=True)
+    response = requests.get(url=url, stream=True)
     response.raise_for_status()
     total = int(response.headers.get("content-length", 0))
+    chunk_size = 1024 * 1000  # 1 MiB
 
     # create a progress bar
     bar = tqdm(
@@ -45,7 +46,7 @@ def download_file_url(url: str, destination: Path) -> None:
     with destination.open("wb") as f:
         with requests.get(url, stream=True) as r:
             r.raise_for_status()
-            for chunk in r.iter_content(chunk_size=1024 * 1000):
+            for chunk in r.iter_content(chunk_size=chunk_size):
                 size = f.write(chunk)
                 bar.update(size)
         bar.close()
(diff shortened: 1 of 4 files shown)


src/refiners/conversion/utils.py
```diff
@@ -63,8 +63,8 @@ def __init__(
         self,
         repo_id: str,
         filename: str,
-        expected_sha256: str,
         revision: str = "main",
+        expected_sha256: str | None = None,
```
deltheil, line 67: Let's tweak the docstring

src/refiners/conversion/utils.py
```diff
@@ -135,6 +135,7 @@ def local_hash(self) -> str:
 
     def check_local_hash(self) -> bool:
         """Check if the sha256 hash of the file in the local hub is correct."""
+        assert self.expected_sha256 is not None, f"{self.repo_id}/{self.filename} has no expected sha256 hash"
         if self.expected_sha256 != self.local_hash:
```
deltheil, line 139: nit: I think `local_hash` could be renamed into `local_sha256_hash` (to echo `hf_sha256_hash`)

deltheil (commented):
Please take a look

src/refiners/conversion/utils.py
```diff
@@ -127,24 +129,32 @@ def local_path(self) -> Path:
         return self.hub_location() / self.repo_id / self.filename
 
     @property
-    def local_hash(self) -> str:
+    def local_sh256_hash(self) -> str:
```
deltheil, line 132: This should be `local_sha256_hash`

src/refiners/conversion/utils.py
```diff
@@ -73,14 +74,14 @@ def __init__(
             repo_id: The repository identifier on the hub.
             filename: The filename of the file in the repository.
             revision: The revision of the file on the hf hub.
-            expected_sha256: The sha256 hash of the file.
+            expected_sha256: The sha256 hash of the file, to optionally check against the local or remote hash.
```
deltheil, line 77:
```suggestion
            expected_sha256: The sha256 hash of the file, to optionally (but strongly recommended) check against the local or remote hash.
```

deltheil (approved):

PR #478, opened 2024-12-20
torch 2.5

catwell:


src/refiners/training_utils/clock.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         self,
         training_duration: TimeValue,
         gradient_accumulation: Step,
-        lr_scheduler_interval: TimeValue,
+        lr_scheduler_interval: Iteration | Epoch,
```
deltheil, line 25: Perhaps add an inline comment for Epoch (FIXME) - WDYT?

deltheil (commented):
PTAL - one minor remark

src/refiners/training_utils/clock.py
catwell: Actually this attribute is useless, I removed it (and added the comment elsewhere).

deltheil (approved):

